PR #18, opened 2018-03-14
Updated load_nbarx and added load_sentinel functions

CEKrause:


algorithms/DEADataHandling.py
```diff
@@ -88,3 +91,52 @@ def load_nbarx(sensor, query, bands_of_interest, product = 'nbart'):
...
+    
+    inputs
+    dc - handle for the Datacube to import from. This allows you to also use dev environments
+	 if that have been imported into the environment.
+    product - string containing the name of the sentinel product to load
+    query - A dict containing the query bounds. Can include lat/lon, time etc
+
+    optional:
+    bands_of_interest - List of strings containing the bands to be read in. Options
+                       'red', 'green', 'blue', 'nir', 'swir1', 'swir2'
```
robbibt: These aren't the only sentinel bands... could update with full list of bands tabled in "GettingStartedWithSentinel2", or direct users to call "dc.list_measurements()" given there's many of them

robbibt: Whoops, didn't realise leaving a comment would prevent me from accepting this... I think @BexDunn should still be able to close it!

CEKrause:
Updated the metadata to fix the references to the bands in the Sentinel function.
@BexDunn The function already has `import numpy as np` at the top, unless I am missing a spot?

rjdunn:
Clearly I can't read... Don't mind me. Sorry!

PR #55, opened 2018-04-17
Add reproject_to_template function to SpatialTools

robbibt:
Added new reproject_to_template function for reprojecting a raster into the extent, cell size, projection and dimensions of another raster


BexDunn:
Nice!

PR #61, opened 2018-05-04
Shuffle

CEKrause:
Restructure of the DEA Notebooks repo.


robbibt:
I had to head home and probably won't be back in today, but one thing I'd consider is adding numbers to the front of the directories to keep them in a logical order, since getting a logical workflow was one of the things we wanted to get out of the shuffle (they currently default to alphabetical and are in a bit of a jumble). Something like this could work:

```
1_Getting_started
2_DEA_datasets
```

robbibt:
And a super minor thing: `Data_output` might be more consistent with the noun-verb phrasing of `Image_classification`/`Index_calculation`/`Composite_generation` than the current `Outputting_data`

robbibt:
Looks amazing!

PR #87, opened 2018-05-17
Fix plotting module fail by adding warnings import

robbibt:


robbibt:
@asivapra The function tries to plot the image with spatial axes so you can then plot things like shapefiles over the top later... the warning should only appear if the xarray you're plotting doesn't have spatial data. What do you get if you run:

`xarray_dataset.extent.boundingbox`?

robbibt:
I've updated the warning to a print statement... it's not really something "wrong" per-se: the user just needs to be notified so they understand what the function is doing

PR #88, opened 2018-05-18
added two functions (polygon and netcdf) to scripts and added geomedi…

BexDunn:
Uploading the geomedian composite dashboard notebook. (Note I'll re-run it and update with a better composite, but all the code and the doco should be about right - please have a check through and let me know what you think.

I've added two functions to DEADatahandling - this notebook uses a function to make loading polygons from shapefiles cleaner, and one to write a composite with no time dimension to NetCDF. The NetCDF writer will need updating at some stage to work for multiple timesteps.

Additionally, we might want to consider removing our geotiff writing scripts as the dea module geotiff writer has included all relevant functionality and works well! This notebook uses it :)


robbibt:
This is super cool! Everything looks good, only thing was in the `write_your_netcdf` function, can we safely remove the following commented out lines? If so, I think I can do it here without you having to re-push the files

```
#dataset.attrs['affine'] =affine
#dataset.dataset_name.attrs['crs'] = crs
```

robbibt:
Oh, and very much agree with adopting any new dea module functions and retiring ours as soon they appear... one central tested codebase is far better than lots of functions that do the same thing.

PR #95, opened 2018-05-18
Final commits of 3 documents

asivapra:
The documentations on LandSats, PQ and FC products.


robbibt:
One thing I'm not seeing in here is anything about the `ga_pq_fuser` function from `from datacube.helpers import ga_pq_fuser` that is usually used inside `dc.load`. It's one of the more confusing parts of loading in PQ data, so it would be great to have a definitive explaination of what it is and how to use it. For example:

```
dc.load(product = "ls8_pq_albers",
             group_by = 'solar_day',
             fuse_func=ga_pq_fuser,
             **pq_query)
```

robbibt:
One other thing that always confuses me: does setting e.g. `nir_saturated=False` inside `masking.make_mask` mean you want to mask out all `nir_saturated` pixels (i.e. False as in you want to get rid of them), or that you want to _not_ mask them out (i.e. False as in you don't want to mask them)?

robbibt:
OK, one more small thing: can the notebooks be renamed to `GettingStartedWithLandsat5-7-8.ipynb` and `Landsat5-7-8-PQ.ipynb`? The correct spelling is without the "s" and with no capital in the middle.

asivapra:
@robbibt 
Robbi, do you have any info/link to learn more about 'ga_pq_fuser' ? the only info I got is this:

**fuse_func** – Function used to fuse/combine/reduce data with the group_by parameter. By default, data is simply copied over the top of each other, in a relatively undefined manner. This function can perform a specific combining step, eg. for combining GA PQ data.

robbibt:
@asivapra Unfortunately not... my basic understanding is that it lets you merge PQ imagery in a way that preserves the different bits without just copying one layer over the top of the other, but that's as far as I've gotten with it.

asivapra:
@robbibt 
How does this read? does it make sense and/or is the correct explanation?

"In the load command below, we are getting the data from 'ls8_nbar_albers' for the specified date range and geo-coordinates. If there are more than one picture on a day, they are combined into one by the group_by='solar_day' parameter. 

ds = dc.load(product='ls8_nbar_albers', fuse_func=ga_pq_fuser, group_by='solar_day', **query)

During this process of making a composite picture, we need to make sure that the data integrity is maintained. By default, data is simply copied over the top of each other, in a relatively undefined manner. This 'fuse_func' can perform a specific combining step, to avoid duplicate data from scene overlaps. It is used here so that the PQ masking (see later) will be more accurate."

asivapra:
@robbibt 
Also, this:

"Based on the property of the pixels, as set above and described in the table below, we can selectively mask some to improve the visualisation."

"    Saturated: means this band has exceeded the dynamic range of the sensor. The data cannot be trusted.
        When it is specified as e.g. blue_saturated=False, it asks to show the pixel only if it is not saturated. If set as True, it will mask ALL pixels that are not saturated for the Blue band.
        The purpose of this mask is to find areas where the reflectance has exceeded the dynamic range of the sensors. e.g. a bright light from reflections on water surface may saturate the pixels.
        Generally, setting any bands as 'saturated=True' will resulty in a black rectangle instead of showing any picture.
    algorithms: to detect clouds which obstruct and/or distort the data.
        ACCA: Automated Cloud Cover Assessment
        Fmask: Function of mask
            ACCA is already widely used within the remote sensing community; it is fast and relatively accurate. Fmask on the other hand is newer, but is rapidly becoming more established, and can provide a more accurate cloud mask than ACCA in certain cloud environments.
            As in the case of the saturated mask, a mask for cloud will do the same. e.g. 'cloud_acca=no_cloud' means that the pixels under a cloud are masked, but all others are displayed. If set as True, only those under cloud are shown and others masked. For some unknown reason, you must set both algorithms to see an effcet. In other words, if set cloud_acca='cloud' and cloud_fmask='cloud', then you will see only the clouds and nothing around it.

The different masks and their bit values can be seen in the table below. Note: The setting of bits as True/False are in reverse order, such that True=0 and False=1, which is confusing to the first-time user. But, if you stop to think about it, it makes sense that 'saturated=False' is the same as 'not_saturated=True'. It is unknown why the mask flags are not named as e.g. 'blue_not_saturated=True' instead of 'blue_saturated=False'."

robbibt:
This sounds great! I would just remove the final sentence ("It is unknown why the mask flags are not named as e.g. 'blue_not_saturated=True' instead of 'blue_saturated=False'."")... the rest of your description makes plenty of sense.

I don't mind if you delete this pull request and do another one... we will still be able to access it via the pull requests tab above. Also, if you include something like ("This replaces pull request #95 " in your commit description, Github should add in a link!

PR #97, opened 2018-05-21
Final commits of 3 Landsats Documents - v 1.2

asivapra:
Added the feedbacks from Robbi to the PQ doc. Changed the names of all files.


robbibt:
I'll let @CEKrause have the final say, but everything looks pretty good to me! (see #95 for comments on changes)

PR #107, opened 2018-05-23
Mineral Indices Calculation notebook and band index script changes

BexDunn:
Added a notebook to calculate mineral indices. All but one of the band ratios has a link to the reference - couldn't find one for NIR/SWIR1 Clay Ratio . Also changed wording of the print statements in the band index script, but didn't add the other indices as they seem to be less widely used and are easy to calculate (as in the script in my notebook).
Bex


robbibt:
It looks great! Another thing we might want to think about in the future is whether we really need the band indices to be in a separate script/module... all the functions in there seem like pretty natural fits for `DEADataHandling` if we wanted to simplify things. :)

robbibt:
@BexDunn It looks great! One thing we might want to think about in the future is whether `band_indices` should be its own separate script/module... most of the functions in there could probably work just as well in the `DEADataHandling` script if we wanted to simplify things.

BexDunn:
https://nbviewer.jupyter.org/github/GeoscienceAustralia/dea-notebooks/blob/e746fd804c6d8b11d4da053381b1c52b1f857e14/Index_calculation/Mineral_Indices_Calculation.ipynb 

The equations render fine on nbviewer!

PR #115, opened 2018-05-24
fixes to nodatavalue filtration in 3 fns

BexDunn:
Andrew, can you please check that we are now handling nodatavalues correctly for sentinel? We've realised that we were not handling them correctly for landsat as we had assumed only NBART had -999.0s

This pull request should fix this in our 3 loading functions (load_nbarx, laod_sentinel, load_clearlandsat, )
Thanks!

### DIFF
--- a/Scripts/DEADataHandling.py
+++ b/Scripts/DEADataHandling.py
@@ -13,8 +13,7 @@ Available functions:
     write_your_netcdf
 
 Last modified: May 2018
-Author: Claire Krause
-Modified by: Robbi Bishop-Taylor, Bex Dunn
+Authors: Claire Krause, Robbi Bishop-Taylor, Bex Dunn
 
 '''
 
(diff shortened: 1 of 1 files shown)


Scripts/DEADataHandling.py
```diff
@@ -172,7 +159,8 @@ def load_sentinel(dc, product, query, filter_cloud=True, **bands_of_interest):
         print('loaded {}'.format(product))
         if filter_cloud:
             print('making mask')
-            clear_pixels = np.logical_and(ds.pixel_quality != 2, ds.pixel_quality != 3)
+            clear_pixels = np.logical_and(ds.pixel_quality != 0, ds.pixel_quality != 2, 
+                                          ds.pixel_quality != 3)
```
andrewdhicks: This doesn't explain what is being masked out.
See https://nbviewer.jupyter.org/github/opendatacube/datacube-core/blob/develop/examples/notebooks/Datacube_Summary.ipynb#Masking-out-cloud

Scripts/DEADataHandling.py
```diff
@@ -110,22 +109,10 @@ def load_nbarx(dc, sensor, query, product='nbart', bands_of_interest='', filter_
             ds.attrs['crs'] = crs
             ds.attrs['affine'] = affine
 
-        # If product is NBAR-T, also correct terrain by replacing -999.0 with nan
-        if product == 'nbart':
+        # Replace nodata values with nans
 
-            print('Masked {} with {} and filtered terrain'.format(product_name,
-                                                                  mask_product))
             ds = ds.where(ds != -999.0)
```
andrewdhicks: Hard-coding this value isn't the best. The value is stored as a attr on the data array.
Check out the `mask_invalid_data` functions here: http://datacube-core.readthedocs.io/en/latest/dev/api/masking.html

Scripts/DEADataHandling.py
BexDunn: probably because I don't understand the sentinel masking

PR #116, opened 2018-05-24
Add WOfS notebook and script to deanotebooks

BexDunn:
Uploading a WOfS notebook and script to run WOfS


robbibt:
This looks super awesome. Is it possible to move it to `Image_classification` though? It's definitely more of an image classification challenge, and the decision tree approach of WOfS goes nicely with the existing random forest classification notebook.

The .rst file needs to be updated too :)

robbibt:
Wow, "requesting changes" produces an excessively angry looking "X"!

PR #126, opened 2018-05-31
FC notebook changes and README

BexDunn:
Made some important changes to the FC notebook:
-previous version of notebook referenced different datasets (QLD and MODIS) so explanation of errors has been written for the version we have as per our product description.

- Added a fc load function to DEADataHandling.py as part of loadclearlandsat, so it is cleaner and filters out nodata values. Previously nodata values were still in data.

-Wrote capability to write to netcdf and geotiff into notebook, for either single or multiple timesteps.

-updated references to correct refs for the FC25 dataset we have and clarified a few things.

#TODO add in wofs filtration for water bodies.


BexDunn:
note - first commit is a stuff up, but 4th commit fixes the stuff up

robbibt:
"Digital Number 10000 = 100%"... is that true? When I load them in they come out as values from "0" to "100"

PR #127, opened 2018-05-31
Add function for calculating zonal statistics

cbur24:
As 'rasterstats' package is not a default DEA module, zonal_timeseries functions is commented out.  Do not accept this pull request until rasterstats has been installed and function has been uncommented.
@BexDunn @CEKrause

### DIFF
--- a/Scripts/DEADataHandling.py
+++ b/Scripts/DEADataHandling.py
@@ -11,6 +11,7 @@ Available functions:
     dataset_to_geotiff
     open_polygon_from_shapefile
     write_your_netcdf
+    zonal_timeseries
 
 Last modified: May 2018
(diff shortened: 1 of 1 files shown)


BexDunn:
You could just include a requirements section like @CEKrause  does:

We also require the rasterstats python library. This is not currently part of the DEA library and so needs to be installed by the user. To do this, type pip install rasterstats --user from a terminal window on the VDI. The dependencies of this library do not interfere with the DEA libraries.

BexDunn:
Actually, it does say in the function that you need rasterstats, so maybe just include how to install it in the function info. Can you please uncomment the function because it's hard to do from within the pull request (unless you know how to do it more easily than by adding another commit)

PR #132, opened 2018-06-01
added new notebook for glint correction

SnowyH:


robbibt:
@BexDunn I think it might just be that it's a big notebook so Github sometimes doesn't load it properly... it worked OK when I tried this morning:
capture

PR #133, opened 2018-06-03
Update Landsat and PQ notebooks to close #118, #109 & #73 

### LINKED ISSUE #73: Better organisation for loose images in dea-notebooks/DEA_datasets
The new Landsat notebooks come with some extra local images that are currently located at `dea-notebooks/DEA_datasets`. I can see that some of these "GettingStarted..." notebooks might need local files like this, but I think we might want to organise them a bit better. Perhaps they could all be stored in a folder in that directory? Something like:

`dea-notebooks/DEA_datasets/images`

@asivapra @CEKrause

### LINKED ISSUE #109: Update GettingStarted notebooks to use DEA functions instead of hardcoding
Currently some of the example notebooks (e.g. "GettingStartedWithLandsat5-6-7") use older versions of the DEA functions that have been hard-coded into the notebook. 

Instead, we should use the existing DEA module functions.

### LINKED ISSUE #118: Need to add -999 as nodata for NBAR in GettingStartedWithLandsat5-7-8 notebook
The `GettingStartedWithLandsat5-7-8.ipynb` notebook currently includes "Note that the NBAR-T data inserts a missing value (-999) into the dataset to denote a terrain shadow. This should be converted to a NaN value before using this data to avoid treating it as a valid value."

However, NBAR also includes -999 values for nodata. This should be covered both in the dataset description **and** in the actual code examples.

robbibt:
I've added a chunk of extra description specific to DEA's Landsat archive (rather than just generic Landsat), and some extra examples of basic data manipulation (e.g. band index calculation).

For PQ, I've fixed up the description of PQ masking to include a step-by-step example that steps through buildings a progressively more complete/complex mask.

Renamed Intro to HLTC and ITEM notebooks.

Moved image files into a subfolder to prevent folder from getting filled with loose files.


robbibt:
Closes #118, #109, #73

robbibt:
(Images in Intro to Landsat 5, 7, 8 won't load until after the pull request is accepted, but they render on this staging branch so everything should be good!)

robbibt:
Closes #109 
Closes #73

PR #138, opened 2018-06-05
Update notebook to work with latest versions of animation functions

robbibt:


robbibt:
Ahh, overwrote some earlier changes. Will fix!

robbibt:
OK, ready to go now

PR #164, opened 2018-06-27
Hovmoller TCI notebooks

BexDunn:
Multi site and single site hovmoller workflow notebooks, demonstrating how to load rainfall data and plot a hovmoller with TCI, monthly and yearly  rainfall, and thresholded TCI.


robbibt:
So cool! Just need an index update and it's good to go!

BexDunn:
thanks Robbie!

PR #170, opened 2018-06-29
Fractional cover quantiles notebook

BexDunn:
Fractional cover quantiles notebook. last of the basic notebooks on my list


BexDunn:
I will fix  :)

PR #183, opened 2018-08-03
Add the getting started notebooks from the DEA repo

omad:
- Also fix some restructured text issues

See new structure and example output at https://damien-dea-docs.netlify.com


omad:
*updated*

Renamed all the directories so that they're in the right order in the Table of Contents.

Have also updated the preview.

PR #197, opened 2018-08-29
Update links and README

CEKrause:


BexDunn:
You're getting the weird concatenate error from the current version of xarray and it's printing out a whole lot of nasty error statements: may be try re-running it with the older version of dea>?

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

CEKrause:
Thanks @BexDunn. Try now.

PR #199, opened 2018-08-29
Makesuppfolder

BexDunn:
I have moved all the images and data of which I am aware out of individual folders and into two new folders within the new Supplementary_Data folder. If you can't spell supplementary I'm sorry, please use tab complete ;)


robbibt:
Can I be a horribly annoying person and ask that we rename the directory to `Supplementary_data` without the second capital? That way it matches the format of the other folders (i.e. `08_Outputting_data`)...

BexDunn:
done!

PR #205, opened 2018-08-30
Update the READMEs with additional info and caveats

CEKrause:


CEKrause:
@BexDunn syntax error now fixed! :)

PR #206, opened 2018-08-30
Add small note at top of non-reproducible grad notebooks

robbibt:


BexDunn:
I suggest additionally removing the color formatting, as it doesn't display on github

PR #208, opened 2018-08-30
Update FindingDams notebook

robbibt:
Updated reference to scripts directory
Fixed minor bug where wet observations were summed across bitflags rather than booleans


robbibt:
Got rid of the '100' multiplier so that the '0.05' proportion test works correctly (i.e. doesn't compare percents with proportions)

BexDunn:
-wofs plot too?

PR #210, opened 2018-09-11
adding the doctesting ability to the end of the file. Run {filename}.…

BexDunn:
This adds doctest (apparently the base level unit testing thing) to the end of our modules.
--This should only run when you run the script.
 Run {filename}.py to test the file
--I don't think this breaks anything.

**You can put your examples in the docstring and doctest will run them as tests!!!**

```
# If the module is being run, not being imported! 
# to do this, do the following
# run {modulename}.py)

if __name__=='__main__':
#print that we are running the testing
    print('Testing..')
#import doctest to test our module for documentation
    import doctest
    doctest.testmod()
    print('Testing done')
```

https://docs.python.org/3/library/doctest.html#module-doctest

### DIFF
--- a/10_Scripts/BandIndices.py
+++ b/10_Scripts/BandIndices.py
@@ -191,4 +191,15 @@ def tasseled_cap(sensor_data, tc_bands=['greenness', 'brightness', 'wetness'],
 
     return output_array
        
-       
+
+# If the module is being run, not being imported! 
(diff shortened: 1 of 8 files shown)


robbibt:
Fingers crossed nothing breaks! It shouldn't unless we run the script from command line

PR #216, opened 2018-09-13
Claire k sentinel2

CEKrause:


robbibt:
This is super awesome, but to me `Composite generation` doesn't seem like the best location for it... Perhaps `Outputting data` as the best out of a bunch of not perfect options?

PR #224, opened 2018-09-19
Add contour extraction function and notebook

robbibt:
Contour extraction function which is used to extract NIDEM waterline contours... might be useful for other things too. Function has doctests

Also added a notebook demoing functionality with examples for DEMs and waterline extraction


robbibt:
Sorry about the messy commit history! TThe latest version of the notebook is at 119d6ae

seanchua:
Nice work, note the ability to create contours from non-elevation based data. Could make it easier to grok some complex datasets.

PR #229, opened 2018-09-24
Update scripts to cope with new Sentinel. Add new phenology notebook

CEKrause:


robbibt:
Such a cool notebook!

robbibt:
One comment on the script: the SWIR1 version of NDWI is usually called the [Modified Normalised Difference Water Index](https://www.researchgate.net/publication/232724072_Modification_of_Normalized_Difference_Water_Index_NDWI_to_Enhance_Open_Water_Features_in_Remotely_Sensed_Imagery)... if you can be bothered renaming it it might be good to make it easier for other people to find in the future, but it's not a big deal

PR #233, opened 2018-10-26
Uncommenting zonal timeseries as rasterstats is now installed

cbur24:


robbibt:
Completely forgot to do this - good catch! 
@cbur24 You should post this on `look-i-did-a-thing` even though it's from a while ago now that people can actually use it

PR #234, opened 2018-10-28
script to load rainfall without offset

BexDunn:
This script can be run to load the rainfall grids without the spatial offset problem.


robbibt:
Looks great! Can you add the new script and function description to the `10_Scripts` readme file?

https://github.com/GeoscienceAustralia/dea-notebooks/tree/master/10_Scripts

PR #235, opened 2018-10-28
cumulative mass residual notebook and readme

BexDunn:
Notebook to calculate cumulative mass residual rainfall over a given epoch on a rainfall grid cell basis.


robbibt:
Hard to tell on github, but there might be something funny going on with the tags section:
capture

Might also be good to add in the same header structure we've been using in the other notebooks, eg

```
Before you run this notebook: You need to run the following commands from the command line prior to launching jupyter notebook from the same terminal so that the required libraries and paths are set.

module use /g/data/v10/public/modules/modulefiles

module load dea

```

BexDunn:
should be fixed now :)

PR #271, opened 2019-02-28
Sentinel 2 scenes to multiple geotiff files demo notebook

BexDunn:
This notebook gets S2 data for a polygon of choice and writes to multiple geotiffs.


BexDunn:
Yes - the grand master plan is to update the function  with type forcing and multiple timesteps :grinning:

PR #277, opened 2019-03-05
Update remaining Scripts references in DEAPlotting

robbibt:
Also remove redundant ` kwargs.update({'aspect':aspect}) ` (this is automatically passed in without having to set the dict)


robbibt:
Woooooooah hang on, don't accept this!

PR #279, opened 2019-03-06
Bex staging

BexDunn:
Added the landsat to geotiff multiple scenes notebook I wrote for the hackathon

### DIFF
--- a/08_Outputting_data/LandsatScenestoGeoTIFF.ipynb
+++ b/08_Outputting_data/LandsatScenestoGeoTIFF.ipynb
@@ -1,0 +1,471 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2019-02-24T21:39:57.612969Z",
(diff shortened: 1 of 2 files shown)


robbibt:
@BexDunn Looks awesome - there's just some content in the documentation that refers to TCI stuff which I don't think is in the actual notebook:
```
Background: Data from the Landsat 5,7 and 8 satellite missions are accessible through Digital Earth Australia (DEA).

The Tasseled Cap Index (TCI) is a method of reducing 6 bands of satellite data (BLUE, GREEN, RED, NIR, SWIR1, SWIR2) to 3 bands (Brightness, Greenness, Wetness) using a Principal Components Analysis and Procrustes' Rotation (Roberts et al 2018). This notebook uses the published coefficients of Crist 1985 as applied to Digital Earth Australia's Landsat satellite data.This notebook produces the wetness index.

What does this notebook do?: This notebook takes a supplied shapefile of a polygon and queries the datacube Landsat surface reflectance data. It calculates tasselled cap wetness. The results and input surface reflectance are output as geotiffs for each individual scene.
```

PR #291, opened 2019-05-31
Changes to FC notebook

erin-telfer:
and addition of png and shapefile


BexDunn:
looks great! @erin-telfer

PR #304, opened 2019-07-03
example results of image segmentation

cbur24:


cbur24:
Thanks, Bex. Ah okay, wasn't sure if that was good practice or not.

PR #320, opened 2019-08-02
fixed bug with PQ attributed in load_clearlandsat

richardt94:
load_clearlandsat in DEADataHandling.py had a bug where running xr.concat() on datasets from multiple satellites (e.g. ls7 and ls8) would wipe the attributes from combined_ds.pixelquality and change its dtype to float64. This means the definition of the bit flags in the pixel quality product is lost and bitwise operations can't be performed on the resulting data values (as they are floats).  This change fixes this issue by casting the combined dataset back to int and setting its pixel quality attributes to those of the first dataset being concatenated. Possibly not the best way of fixing but does the job.


richardt94:
Discovered that it's actually the mask_invalid_data flag that casts the dtype to float, not the xr.concat() method. Therefore I added a warning when the user attempts to use this flag with the pixel quality product, as casting the bit flags to float is generally a bad idea.

robbibt:
I'd never thought to use `load_clearlandsat` to load PQ before... this looks like a good catch! My only comment would be that both `mask_invalid_data` and `mask_pixel_quality` cast to floats because they insert NaNs... I'd probably update the following to give the warning when _either_ `mask_invalid_data` or `mask_pixel_quality` is used in combination with `product == 'pq'`?

richardt94:
> 
> 
> I'd never thought to use `load_clearlandsat` to load PQ before... this looks like a good catch! My only comment would be that both `mask_invalid_data` and `mask_pixel_quality` cast to floats because they insert NaNs... I'd probably update the following to give the warning when _either_ `mask_invalid_data` or `mask_pixel_quality` is used in combination with `product == 'pq'`?

Updated the warning case and message to reflect the effect from mask_pixel_quality as well.

PR #325, opened 2019-08-12
Add new notebook for modelling intertidal elevation a la NIDEM

robbibt:
New notebook to demo to Qld gov at workshop this month.

For now, we're not worrying about tags and index files for `dea-sandbox` notebooks.

### DIFF
--- a/DEA_sandbox/Case_Studies/Intertidal_Elevation.ipynb
+++ b/DEA_sandbox/Case_Studies/Intertidal_Elevation.ipynb
@@ -52,7 +52,7 @@
     "import numpy as np\n",
     "import matplotlib.pyplot as plt\n",
     "from utils import waterline_funcs\n",
-    "from utils.display import display_map\n"
+    "from utils.display import display_map"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


caitlinadams:
Hey Robbi,

This is looking great! I think it's a fantastic demonstration of how you generate an intertidal DEM. I've made a few small changes re. code style that I'll commit to this branch shortly. 

I noticed that prior to making the faceted plot (included below), you mentioned that tide_interval = 1 corresponds to low tide, and tide_interval = 9 corresponds to high tide. However, to me, it looks like the largest amount of land area (what I'd take to be low tide) occurs in the tide_interval = 9 image, and the lowest amount of land area (high tide) occurs at tide_interval = 1. 
Screen Shot 2019-08-12 at 2 41 15 pm
Is this the correct interpretation of the plot, or am I misunderstanding something?

robbibt:
@caitlinadams  Ah, I think that's just a problem with visualisation: high values (red) are currently water, low values (blue) are land, which is super counterintuitive. The colour scale should be reversed so blue == water!

caitlinadams:
Ah right, I had incorrectly assumed the mapping was in reverse. Should be an easy fix by changing the order of the colour map.

PR #329, opened 2019-08-15
adding explicit dask functionality to load_clearlandsat

cbur24:
Two new variables to 'load_clearlandsat':
- dask_chunking: add dictionary of dask chunks to allow for explicitly specifying spatial and temporal chunking
- lazy_load: adding boolean to allow for optionally delaying the computation of load_clearlandsat - facilitates easy parallel processing with dask


robbibt:
That looks great! Nicely documented too

PR #334, opened 2019-08-27
Update coastal erosion notebook to use OTPS tidal model

robbibt:
We have a tidal model installed on the sandbox now! have updated the `Coastal_Erosion` notebook to use the new tidal model, which theoretically allows users to specify any coastal location for analysis without requiring a pre-canned tidal height file. This should make the notebook a lot more interactive and useful.


robbibt:
Have forced `rgb` to plot the correct aspect, which should be helpful for lots of other applications too. 

For the dark image, this is one of the few areas where `rgb`'s defaults don't look so good due to the bright white sand not playing well with the percent clip. Have added `percentile_stretch` to brighten the image (hope this doesn't make things too complicated).

Have clarified that the area needs to be over water for the tidal model to work in several places in the notebook.

caitlinadams:
The aspect correction looks much better. It's a shame about the rgb defaults not working well for this image, but the percentile_strech seems like a reasonable solution for now. Thanks for adding the additional qualification about the centre needing to be over water. Hopefully, this will help to troubleshoot if people run into issues.

I made a few minor edits to conform function calls to the PEP8 standard we've been using previously. I noticed, however, that your original layout is also PEP8 compliant (see https://www.python.org/dev/peps/pep-0008/#indentation). It might be worth talking through whether there's a preferred convention among people as part of the notebook overhaul.

PR #338, opened 2019-08-30
Update notebook template

caitlinadams:
- move additional information to end of notebook
- add example packages to load, including sourcing the scripts directory
- add information on highlighting parameters to getting started section
- add guideline around using subheadings
- break up markdown paragraphs so each sentence is on a new line to aid git-diff in future

Leaving as a draft PR for now so people can comment and make any further changes.

### DIFF
--- a/DEA-notebooks_template.ipynb
+++ b/DEA-notebooks_template.ipynb
@@ -11,19 +11,10 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "**License:** Code freely available under the [Apache License, Version 2.0](https://www.apache.org/licenses/LICENSE-2.0).\n",
-    "\n",
-    "**Contact:** If you find an issue with this notebook or need assistance, please post a question on the [OpenDataCube slack channel](https://opendatacube.slack.com), ask a question on StackExchange using the [open-data-cube tag](https://gis.stackexchange.com/questions/tagged/open-data-cube), or file an issue on [Github](https://github.com/GeoscienceAustralia/dea-notebooks).\n",
-    "\n",
-    "**Last modified:** August 2019"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
     "### Background\n",
-    "An *optional* overview of the scientific, economic or environmental management issue or challenge being addressed by Digital Earth Australia. For `Getting started` or `Common analysis techniques` notebooks, this may include information about why the particular technique or approach is useful or required. If you need to cite a scientific paper or link to a website, use a persistent DOI link if possible and link in-text (e.g. [Dhu et al. 2017](https://doi.org/10.1080/20964471.2017.1402490))."
+    "An *optional* overview of the scientific, economic or environmental management issue or challenge being addressed by Digital Earth Australia. \n",
+    "For `Beginners_Guide` or `Frequently_Used_Code` notebooks, this may include information about why the particular technique or approach is useful or required. \n",
+    "If you need to cite a scientific paper or link to a website, use a persistent DOI link if possible and link in-text (e.g. [Dhu et al. 2017](https://doi.org/10.1080/20964471.2017.1402490))."
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


robbibt:
This is looking great! Just flagging some quick thoughts for now:
- Whatever naming convention we are happy with (e.g. camel case, snake case etc), we should make sure this notebook's file name follows it
- Need to decide which of OpenDataCube Slack or GIS Stack Exchange is the first place to send people, and have this reflected in the order mentioned in "Contact". Ideally it would be great if this was Stack Exchange, but right now most questions on there are going unanswered so I'm not sure...
- Once we have some code in `Scripts` we should add an example import below `sys.path.append('../Scripts')`, just to make it really obvious how you do it
- In theory I like the 'Use markdown to inform the user of any parameters they'll need' bit, but its current location will separate the explanation from the actual parameters which will need to be after the 'Load packages' section. To really encourage people to keep all non-hard coded parameters up the top (which I think we absolutely should do), I'd say we recommend an optional "Parameters" markdown + code cell directly after the "Load packages" markdown + code cell?

caitlinadams:
Thanks for adding the label and updating the project milestones! 

- I've generally been using camel case with underscores between words, and trying to keep notebook file names to no more than a few words. That would make this file: `DEA_Notebooks_Template.ipynb`. I'd also be happy with `DEA_notebooks_template.ipynb`. Any preferences?
- This is a good point -- I was operating on a "be the change you want to see" approach, but you're right that it might be more appropriate to preference the slack channel for now. Happy to change this.
- Agree that it would be good to show an import example, and maybe even an example of appending and loading from the `Supplementary_Data` folder too?
- You're right -- I missed that I had it before load. I'll move it down and add a title like you mentioned.

robbibt:
Nice! Thanks @caitlinadams .

How do we want to go about clearly marking notebooks as suitable for running on `DEA-sandbox` vs. `NCI` vs. both? As much as I'd love all notebooks to work on both, this probably won't be the case for all, especially at the beginning. 

Would a big bold piece of text directly below the heading be too much? It's the kind of thing that we really want people to be aware of before they start running the notebook and get lots of errors...


> # Descriptive title that follows notebook filename
> **Notebook compatible with the DEA Sandbox environment only** 
> 
> ## Background
> An optional overview of the scientific, economic or environmental management issue or challenge being addressed by Digital Earth Australia. For `Beginners_Guide` or `Frequently_Used_Code` notebooks, this may include information about why the particular technique or approach is useful or required.

caitlinadams:
Hey @robbibt -- I like the additions you made, I just tidied them up a little. One of the things I'd like to add to the style guide is that each markdown sentence should start on a new line. This makes assessing changes in Markdown text between git commits really easy, as each sentence has its own line number and can be directly compared. This is a bit of a weird habit to get into, but it makes assessing changes to text really easy. We can talk about it more though :) 

I also added a comment about needing to make sure the tags cell is in Raw format and removed the empty cell from the end of the notebook.

robbibt:
Yep, absolutely - I think I just realised that the really annoying way that carriage returned lines merge into the one paragraph actually has a useful purpose! I'd be happy to have a sentence explaining this in the first example template markdown cell, e.g. "Make each markdown sentence start on a new line so that future changes can be more easily compared" etc

caitlinadams:
I added it into a section on Style tips, which the author should remove before making a PR for the notebook. We can potentially add other style/formatting information here if we come up with anything else.

robbibt:
Have moved the "choosing location" cell up into the top cell so users only have to delete one cell before publishing. Also included a "Compatible datacube version" cell down the bottom because it was mentioned at the last working group meeting that we should document this... happy to take it out if anyone has a nicer way to do this!

Kirill888:
What about `%matplotlib inline|notebook|widget`? Supporting both `lab` and `notebook` is tough, only `inline` works across both. There is no `notebook_or_widget_depending_on_what_user_is_running` that I'm aware.

caitlinadams:
Hey @Kirill888 -- that's a good suggestion. I'll add a sentence about including any magic commands as part of the import, and keep the example of `%matplotlib inline`. Do you know if there's best practice around running magic commands before or after package imports? I've seen both:

```
%matplotlib inline
import matplotlib.pyplot as plt
```
and
```
import matplotlib.pyplot as plt
%matplotlib inline
```

Kirill888:
@caitlinadams not sure what is considered best practice, but I usually put `%magics` up top of the code cell, so before imports, and usually up top of the notebook. I view `%matplotlib inline` as a "whole notebook configuration option", so prefer to put if first.

PR #353, opened 2019-09-13
Add MSAVI and NDCI to calculate_indices

robbibt:
Plus some minor formatting fixes

### DIFF
--- a/Scripts/dea_bandindices.py
+++ b/Scripts/dea_bandindices.py
@@ -36,36 +36,36 @@ def calculate_indices(ds,
         used as inputs to calculate the selected water index.
     index : str
         A string giving the name of the index to calculate:
-        'NDVI' (Normalised Difference Vegetation Index, Rouse 1973)
-        'EVI' (Enhanced Vegetation Index, Huete 2002)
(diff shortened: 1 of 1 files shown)


robbibt:
@cbur24 I think I agree. I originally had the normalisation hard coded for all indices, but that produced different results for Tasseled Cap than have been produced in the past using unnormalised data (not wrong, just out by 10,000). After talking to Ness and Bex yesterday we decided to have normalisation option but off as default so Tasseled Cap results could be replicated. 

That said, I've now realised that over half of the indices in there are invalid if `normalised=False`, so I think the best approach would be to still have normalisation as an option, but set the default to `True` instead

robbibt:
@cbur24 Changed the default to True. There's also a warning if they set `normalisation=False` and then try to use something like MSAVI:
```
        warnings.warn(f"\nA coefficient-based index ('{index}') normally "
                       "applied to surface reflectance values in the \n"
                       "0.0-1.0 range was applied to values in the 0-10000 "
                       "range. This can produce unexpected results; \nif "
                       "required, resolve this by setting `normalise=True`")
```

PR #356, opened 2019-09-19
Add placeholder dea_classificationtools script

robbibt:
# Proposed changes
An empty script to add functions used to classify remote sensing data (e.g. random forest, decision trees, k-means etc)


robbibt:
@richardt94: Chad mentioned you might have some code that would be good to incorporate into DEA notebooks. I've just created a blank script that would be a good place to add in any reproducible functions that help with classifying DEA data with machine learning or other techniques!

PR #357, opened 2019-09-19
New notebooks for CopHub Workshop

caitlinadams:
This PR adds three new case study notebooks:
- Mining
- Forestry
- Waterbody monitoring

These largely follow the template for the notebooks overhaul, with a few minor changes. This is to keep them in line with the content that's already in the sandbox. For example, I haven't added the tags section at the end of each notebook.

@alexgleith, it would be great if you could review all of these notebooks. The dea_x.py scripts have already been reviewed as part of the notebooks overhaul -- so feel free to look over them or not.
@robbibt, it would be great if you could review the Waterbody Monitoring notebook
@cbur24, it would be great if you could review the Forestry notebook

Happy to take any suggestions for changes!


robbibt:
These look very cool, I'll see if I can have a look at them this afternoon. So just to clarify: these notebooks are being committed to the `master` branch at this stage ready for the CopHub worshop which is why they aren't 100% compliant with the `develop` branch requirements, but the plan is to update them and add them to `develop` (with the required changes) at a later stage?

caitlinadams:
Yeah, that's totally right. I've got more `develop`-branch compliant versions of these that I'll do as PRs as part of the notebook overhaul.

cbur24:
@caitlinadams The forestry notebook looks fantastic; it runs perfectly on the Sandbox and I didn't spot any typos/errors. A couple of thoughts I had about the notebook are below but I don't consider them necessary to address, the notebook is fine how it is.

1. The `time_baseline` parameter description could be further clarified to make it more obvious why we split the time-series into two non-overlapping samples (i.e. the assumptions inherent in the Welch T-test).  

2. For the `time` parameter description, perhaps there should be a comment suggesting a minimum time range of 2 years, to ensure the results of the hypothesis testing make sense. This should prevent a user loading a single years' worth of data and then splitting it by winter/summer and 'discovering' a significant difference purely due to the seasonal cycle.

3. The colour scale on the NDVI plots could be set at `vmin=0` to avoid everything looking green

4. This could be tricky to implement, but producing a zonal time series plot of NDVI over both a deforested and afforested region, with a vertical line showing the separation of the baseline period from the change period, could provide a useful pedagogical tool to understand why significance testing is useful in Australia given the large climate variability we experience.  A time series for a pixel might even be enough, then you could use the display map function to find a lat/lon coord to pass to the `xarray.sel()` method.

caitlinadams:
Thanks for your comments, @cbur24. I think those are all good suggestions and will help clarify different aspects of the notebook. 

Your fourth point is something I'd definitely like to implement, but I don't have the time before we run this workshop. I'd definitely like to try and add it as part of the notebooks overhaul. This would tie in well if we can get the polygon thing to work too -- that way you could plot the line for regions that underwent significant change. I definitely think that something like what you had in the poster would add to the notebook.

robbibt:
The `Waterbody_monitoring` notebook is really fantastic! I've just made some very minor changes do documentation, and simplified some code (e.g. removed some `.values` which didn't seem to be needed, and used `rgb`'s  multiple index functionality. I've sent the notebook to Steve Sagar for any comments on the intro text, but to my knowledge it looks like it represents the science very well.

My only bigger picture comment was that there were a bunch of cells from `Compute band indices` to `Combine the data ...` that give the user no feedback (e.g. that do something, but don't plot or print a result). In future versions it might be good to plot some examples of the outputs in these cells so the user knows that their code is having an effect.

DEA_sandbox/Case_Studies/Forestry.ipynb
```diff
@@ -107,7 +107,7 @@
     "### Analysis parameters\n",
     "\n",
     "The following cell sets the parameters, which define the area of interest and the length of time to conduct the analysis over.\n",
-    "There is also a parameter to define how the data is split in time; this split yields two samples, such that the samples can be compared to detect change.\n",
+    "There is also a parameter to define how the data is split in time; thr split yields two non-overlapping samples, which is a requirement of the hypothesis test we want to run (more detail below).\n",
```
robbibt: Typo

caitlinadams: Well spotted, thanks!

PR #359, opened 2019-09-19
Fix load_clearlandsat casting to float64

robbibt:
By explicitely casting to `float32` before masking operations, we can avoid xarray automatically casting to more memory hungry `float64`.

### DIFF
--- a/10_Scripts/DEADataHandling.py
+++ b/10_Scripts/DEADataHandling.py
@@ -247,10 +247,10 @@ def load_clearlandsat(dc, query, sensors=('ls5', 'ls7', 'ls8'), product='nbart',
         filtered images may still contain up to 1% poor quality pixels. The default of False simply returns the
         resulting observations without masking out these pixels; True masks them out and sets them to NaN using the
         pixel quality mask, but has the side effect of changing the data type of the output arrays from int16 to
-        float64 which can cause memory issues. To reduce memory usage, set to False.
+        float32 which can cause memory issues. To reduce memory usage, set to False.
     mask_invalid_data : bool, optional
         An optional boolean indicating whether invalid -999 nodata values should be replaced with NaN. Defaults to
-        True; this has the side effect of changing the data type of the output arrays from int16 to float64 which
+        True; this has the side effect of changing the data type of the output arrays from int16 to float32 which
         can cause memory issues. To reduce memory usage, set to False.
     ls7_slc_off : bool, optional
         An optional boolean indicating whether to include data from after the Landsat 7 SLC failure (i.e. SLC-off).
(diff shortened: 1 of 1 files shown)


10_Scripts/DEADataHandling.py
```diff
@@ -451,7 +451,8 @@ def load_clearlandsat(dc, query, sensors=('ls5', 'ls7', 'ls8'), product='nbart',
         # Optionally filter to replace no data values with nans
         if mask_invalid_data:
 
-            print('    Replacing invalid -999 values with NaN (data will be coerced to float64)')
+            print('    Replacing invalid -999 values with NaN (data will be coerced to float32)')
```
Kirill888: alignment is off?

10_Scripts/DEADataHandling.py
```diff
@@ -406,7 +406,7 @@ def load_clearlandsat(dc, query, sensors=('ls5', 'ls7', 'ls8'), product='nbart',
 
                     # Optionally apply pixel quality mask to all observations that were not dropped in previous step
                     if mask_pixel_quality:
-                        filtered = filtered.where(good_quality)
+                        filtered = filtered.astype(np.float32).where(good_quality)
```
Kirill888: should we make this configurable? maybe like `masked_dtype='float32'` with ability to overwrite with `float64|16` by users who need more control over memory/precision trade of.

10_Scripts/DEADataHandling.py
robbibt: How do you tell this? It seems to look OK to me...

Kirill888: maybe it's just github diff, I'll checkout code locally

10_Scripts/DEADataHandling.py
robbibt: I agree, but I think the effort should be moved to `load_ard` rather than this function, as we will be retiring this one very soon. I'll create an issue for `load_ard` though

Kirill888: ah,sure didn't realize this was deprecated

robbibt: Yep, this is the equivelent pull request for `load_ard`, this one was a short term fix so Juan Guerschman from CSIRO could run his analyses: https://github.com/GeoscienceAustralia/dea-notebooks/pull/358

10_Scripts/DEADataHandling.py
Kirill888: @robbibt there are tab characters on `combined_ds` line, we should really use spaces everywhere in python

robbibt: Ah, weird, I thought I had all my editors set up to avoid that. Will have a look and replace them with spaces

robbibt: Fixed the tab issue

PR #362, opened 2019-09-20
Adding temporary info to `develop` readme pages

robbibt:
Adding some temporary information to the readme pages for each directory on `develop` to guide the creation of notebooks during the `dea-notebook` refresh

### DIFF
--- a/DEA_datasets/README.rst
+++ b/DEA_datasets/README.rst
@@ -1,4 +1,14 @@
 DEA datasets
 ============
 
-These notebooks are designed to introduce you to the datasets available through Digital Earth Australia.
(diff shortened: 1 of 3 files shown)


caitlinadams:
Hey @robbibt -- the original purpose of these files is to act as the index page for each notebook folder on the dea-docs website. However, I'm happy to have the temporary instructions in them. They may get extended later as we develop the website further. That ok?

robbibt:
@caitlinadams Yep, that's fine. This is just intended to be for the next three weeks so there's some extra guidance for structuring the different kinds of notebooks. We can remove it after when we get everything set up to be rendered on the website.

PR #364, opened 2019-09-24
Same fix for lost CRS attributes for `load_clearlandsat`

robbibt:
Same fix for missing CRS when changing dtype


BexDunn:
looks good to me. thanks

PR #365, opened 2019-09-24
Make linear regression notebook use cloud free data

pkilgour:
Added in cloud free code to the linear recession notebook


robbibt:
Hey @pkilgour , this looks great! Just a few minor things that would be good to fix up:

- The code in `Import modules` is currently pointing to somewhere in your working directory, rather than the correct `'../10_Scripts'`. We also now use a slightly different way of modifying the path, can you please replace this code:
```
# Point this to where you have the algorithms from the dea-notebooks/algorithms saved
#sys.path.append(os.path.abspath('../10_Scripts'))
sys.path.append('/home/547/pxk547/JupyterNotebooks/dea-notebooks/10_Scripts/')
import DEADataHandling, DEAPlotting
```
With this: (as long as you run the notebook from a folder inside DEA Notebooks, you shouldn't need to manually change the path like above :) 
```
# Load custom DEA notebooks functions
sys.path.append('../10_Scripts')
import DEADataHandling, DEAPlotting
```
- There's some extra blank cells at the very bottom of the notebook below "Conclusions"; can you remove them?
- Finally, for future pull requests it's better to give them a descriptive name that describes the main changes you've made. I've edited this one from `Penstemp` to `Make linear regression notebook use cloud free data`.

Apart from that, it looks like a really good fix!

robbibt:
Oh, another thing: we're currently updating all the notebooks on here to work with Collection Upgrade data that is currently being reprocessed for the entire country (this will be the default data once it's ready, and the data we're currently using will be taken down early next year).
I'd love to update this notebook as part of the refresh, but at the moment we only have 30 years of sample data for these areas: http://explorer.sandbox.dea.ga.gov.au/ga_ls8c_ard_3/2019

Are you aware of any interesting areas in those three locations that we could point this notebook to, at least until the full collection upgrade dataset is ready? I guess ideally we'd be after a nice, interesting result like the one currently featured in the notebook

pkilgour:
Ok, thanks for letting me know. No, I don’t know of any interesting areas within those locations as I haven’t worked in any of those areas.
If I have time, I can try and have a look later…
It could be tricky though as I thought I would have an interesting result in the area I’m working in (different to the notebook) and the recession lines were unfortunately the same, so might not be able to….

PR #367, opened 2019-09-25
Notebook s2 ndci

caitlinadams:

### Proposed changes
New real world example notebook about Chlorophyll in waterbodies


robbibt:
@caitlinadams I've made some minor changes above: I rearranged the introduction sections to follow the template a bit better (e.g. moved 'The problem' into 'Background`, and added a 'Description' heading); have a look and see if you're OK with this! I've also made the double quotations consistent, and changed the print statements for the NDCI and NDWI time series to plots so it's a bit easier to interpret. Other than that, I'm just waiting on the NCI to come back online at 2pm today to do a final test of whether it works on the NCI too - will close this off once I do.

caitlinadams:
Thanks for the changes, @robbibt -- great idea to plot the data rather than just display the xarray. I've added some axes labels and changed the plot aspects so all three data plots are visually similar. The setup at the start looks good too, I'll make sure I change the order in the forestry notebook as well. 

I've also fixed the tags as per the new wiki. Would you mind double-checking to see if I missed anything obvious?

PR #368, opened 2019-09-25
Add DEA classification tools to refreshed notebooks

richardt94:

### Proposed changes
Add new script dea_classificationtools.py to interface between xarray (DataArray and Dataset) and sklearn models, including KMeans, PCA, decision trees, etc.

This script is written to be as *general* as possible. Functionality is exposed to the user to turn labelled DataArrays and Datasets into 'flat', unlabelled numpy arrays of pixels, and rebuild a DataArray with spatiotemporal structure from a flat numpy array. This means the user can utilise a scikit-learn model of their choice, and we don't need to write bespoke functions for each model.

This script also contains functions to rasterise polygons of training data to assist with the application of sklearn's supervised classification models to spatial datasets.

Finally, the script contains a few (maybe just one) handy extensions to sklearn's models, such as a hierarchical KMeans model of the type used for the prototype radar wetlands insight tool.


robbibt:
This looks great @richardt94 and @s-m-t-c ! Can't see any issues with the workflow you used. Are you ready for this to be merged in/accepted?

PR #373, opened 2019-09-25
Add Integrating_external_data notebook

robbibt:
### Proposed changes
Added a notebook that demonstrates how to load in external data from a CSV (e.g. tide, river, rainfall gauges), and integrate this with a time series of DEA satellite data.

### Checklist 


robbibt:
@caitlinadams Thanks for the edits! I've added some additional context to the area and band loaded.

The warning message only shows up on the Sandbox (not the NCI), so looks like it's some quirk of the matplotlib/pandas versions installed here. I've followed the recommendation in the warning and registered the converter after we load in the modules; hopefully this will just be a temporary fix and we can remove it later on.

PR #379, opened 2019-09-26
Export to NetCDF

benjimin:
Tested on both VDI and sandbox.


BexDunn:
looks good to me

caitlinadams:
Yes, sorry, I got partway through this today. Will pick up on Monday if that's ok! I've just got a few minor comments about it

robbibt:
Comments addressed and notebook incorporated by #446

robbibt:
Comments addressed and notebook incorporated by #446

Issue #374, opened 2019-09-25
dea_datahandling/load_ard is not preserving attributes


cbur24:
The data-arrays inside the returned xarray dataset from `load_ard` do not retain the crs/geobox attributes.

reproducible example (run on the sandbox):

```
import datacube
import sys
sys.path.append("../Scripts")
from dea_datahandling import load_ard

query = {
    'x': (153.40, 153.45),
    'y': (-28.85, -28.90),
    'time': ('2018-01', '2018-03'),
    'measurements': ['nbart_red', 'nbart_green', 'nbart_blue'],
    'output_crs': 'EPSG:3577',
    'resolution': (-30, 30),
    'group_by': 'solar_day'
}

dc = datacube.Datacube(app="test")
ds = load_ard(dc=dc, **query, products=['ga_ls8c_ard_3'])

#select data array from dataset
da = ds.nbart_red.isel(time=1)
print(da.crs)
```

robbibt:
Suspect that this is being caused by a bug in xarray. Have created an issue here, will see what they come back with: https://github.com/pydata/xarray/issues/3348

PR #381, opened 2019-09-27
Fix load_ard dropping attributes when changing dtype

robbibt:
### Proposed changes 
Due to a possible bug in xarray (see issue https://github.com/pydata/xarray/issues/3348), changing the dtype of an xarray object causes the object to lose attributes from both the main dataset and any data variables. This causes down-stream issues due to missing CRS info, and nodata values defined as attributes on individual variables.

This is the latest attempt at fixing the problem: now trying to apply a custom func to each variable in the dataset that first copies the attribute info, and resets it back after we change `dtype`.

### Closes issues (optional)
- Closes Issue #374


robbibt:
Confirmation that this is caused by an `xarray` bug (we can remove the above fix once it is resolved in `xarray`: https://github.com/pydata/xarray/issues/3348

PR #390, opened 2019-10-01
Image segmentation notebook for Frequently_used_code

cbur24:

### Proposed changes
Object based image analysis is becoming increasingly prevalent in the RS literature, this notebook  demonstrates two methods for conducting image segmentation - a common approach to deriving image-objects.

Notebook is not currently compatible on the Sandbox as rsgislib cannot be installed (requires a conda install). The cells for the shepherd seg have not been run as the NCI is down.


robbibt:
@cbur24 This is a really great notebook. I've tested it on the NCI and everything works great. The biggest change I made was moving the first export geotiff down into the actual KEA section because it seemed like a good idea to keep those workflows together (as you need to export to geotiff to do that segmentation). I also added a simple example at the end showing how to load in the kea file with `xr.open_rasterio`. Let me know if anything doesn't look right, but I'm happy to accept when you are

cbur24:
Looks good, happy to see these changes added to the notebook

PR #391, opened 2019-10-01
Notebook coastalerosion

robbibt:
### Proposed changes
Updated the coastal erosion notebook to use the new template, and moved over all necessary functions required for the notebook.


caitlinadams:
Hey @robbibt, this is looking great. Having a quick read through, I noticed that under the "Compute Modified Normalised Difference Water Index" section, you have a hyperlink to the research paper. What do you think about changing this to be a citation that's hyperlinked, similar to the paper reference in the NDCI notebook?

robbibt:
@caitlinadams  No problem, will update. I've just tried testing the notebook on the NCI and the Collection 3 database seems to be out of action, which means that right now, all the other "NCI compatible" notebooks will cause errors too...

cbur24:
Really nice case study, @robbibt. One quick comment, is it possible to get a color ramp on the shoreline migration plot to make it obvious that the red colours indicate more recent shorelines?  Maybe this is already implemented on the interactive plot (I just read the notebook, I haven't run it), in which case ignore this.

robbibt:
@cbur24 I'd really like a key on that figure too, although custom legends are always a hassle in matplotlib! I'll see if there's an easy option

robbibt:
@caitlinadams @cbur24 Have made both your requested changes (Geopandas made adding a colourbar easy!). If you're happy I think it should be OK to have this merged in even though we can't test it on the NCI - we can come back and test it at the end before we make `develop` the new `master`

PR #394, opened 2019-10-02
Close #393 by calculating aspect on the final two dimensions, rather than excluding the sometimes missing index_dim

### LINKED ISSUE #393: dea_plotting.rgb cannot handle xarrays with no time dimension
If an xarray.dataset has no time dimension, the rgb plotting function fails. 

Reproducible example:

```
import datacube
import datacube_stats
from datacube_stats.statistics import GeoMedian
sys.path.append("../Scripts")
from dea_datahandling import load_ard
from dea_plotting import rgb

# Create a query object
query = {
    'x': (153.35, 153.50),
    'y': (-28.80, -28.95),
    'time': ('2018-01', '2018-06'),
    'measurements': ['nbart_green',
                     'nbart_red',
                     'nbart_blue'],
    'output_crs': 'EPSG:3577',
    'resolution': (-30, 30),
    'group_by': 'solar_day'
}

# Load available data from all three Landsat satellites
ds = load_ard(dc=dc, 
              products=['s2a_ard_granule', 's2b_ard_granule'], 
              **query)

#compute geomedians
geomedian = GeoMedian().compute(ds)

#plot geomedian
rgb(geomedian, index=None,
    percentile_stretch=(0.00, 0.95))
```

Suggest something like:

```
if ds.time is None:
     img = da.plot.imshow(params)
```

robbibt:
Currently the `rgb` function fails when an array with no 'time' dimension was specified. This was accidently caused by the way that the aspect is calculated, which currently tries to exclude the `index_dim` (e.g. `time`) to get the `x` and `y` dims. If there is no `index_dim` in the dataset, this will fail.

Changed to selecting the final two dimensions instead, which works regardless of whether `time` is in the dataset.

### DIFF
--- a/Scripts/dea_plotting.py
+++ b/Scripts/dea_plotting.py
@@ -51,7 +51,7 @@ def rgb(ds,
     This function was designed to work as an easier-to-use wrapper 
     around xarray's `.plot.imshow()` functionality.
     
-    Last modified: August 2019
+    Last modified: October 2019
     
     Parameters
     ----------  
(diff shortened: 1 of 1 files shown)


robbibt:
I'm a bit concerned about how bloated this function is becoming (it was originally meant to be a super simple wrapper), so I might try re-writing it again at some point. It should be a refactor rather than a breaking change though! :)

cbur24:
I tested the fix within the Geomedian notebook and it appears to work well.

PR #395, opened 2019-10-02
Geomedian notebook for Frequently_used_code

cbur24:
### Proposed changes
Notebook implements the data-cube stats geomedian code to allow for image compositing

- [x ] Notebook created using the [DEA-notebooks template](https://github.com/GeoscienceAustralia/dea-notebooks/tree/develop)
- [x ] Remove any unused/empty code cells
- [x ] Include relevant tags in the final notebook cell (refer to the [DEA Tags Index](https://docs.dea.ga.gov.au/genindex.html), and re-use tags if possible)
- [x ] Clear all outputs, run notebook from start to finish, and save the notebook in the state where all cells have been sequentially evaluated


robbibt:
Very nice and simple notebook - I've renamed the file so there's some more context for users who don't know what they're looking for, and changed the colour stretch on the RGB plots. Ready to accept once you mark it as 'ready for review'!

PR #396, opened 2019-10-03
Updates to DIY notebook

caitlinadams:
- fix the issue with axis ticks and replace three_band_image with rgb
- update text throughout
- leave all variables blank for the user to supply, with suggestions in markdown
- use time_step to select the same image in all plots
- add a hint about how to change the colourmap in the NDVI section


caitlinadams:
Thanks Alex!

PR #402, opened 2019-10-07
Intertidal elevation notebook

robbibt:
### Proposed changes
Add an updated version of the intertidal elevation modelling notebook which is refresh-compliant.

Also fixed a small bug in `rgb` which caused RGB plots to have an inverted aspect.

Can't currently test on the NCI because the sample Collection 3 data was deleted.


robbibt:
Looking at this notebook now, it's perhaps a bit too long and complex for its own good. At a later stage, it might be good to wrap some of the functionality into larger analysis functions to reduce the amount of code exposed in the notebook and make things a bit easier to follow for less experienced users.

PR #410, opened 2019-10-09
Notebook wofs bitmask

bellemae:

### Proposed changes
This is a notebook to explain the structure of water observation feature layers WOFLs, and how to use  them for masking.


robbibt:
@erialC-P @bellemae The `wofs_fuser` function has been added to `dea_datahandling`, I think `wofls_fuser` could be as well! https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Scripts/dea_datahandling.py#L464

erialC-P:
In the Introduction, you also mentioned that more options to explore the data would be offered at the end of the notebook but I think these still need to be added (or you can edit your original comment and delete this suggestion!)

katedugdale:
This is an informative notebook, I felt like I could finally understand what a WOFL is after going through it a few times. One thing that I thought would make it easier for first timers to understand is  to add a little bit about what is meant by 'Features' in the very first section:

"Individual water classified images are called Water Observation Feature Layers (WOFLs), where the feature could be something like mountain shadow, ocean, surface water or wet vegetation....."

If the reader has in mind what the feature could be from the start, it makes it a bit easier to follow on the first read-through.

robbibt:
@erialC-P I can't seem to accept and merge this PR... I think you might need to do the final step!

PR #412, opened 2019-10-10
Update to notebook template after refresh sprint

robbibt:
I've made some minor updates to the notebook template based on my experiences using it over the past few weeks. These are completely optional, I just wanted to start the discussion with a pull request so we have something to work from!

1. I've moved the Guidance cell to the very top of the notebook so it's easier to delete
2. I've got rid of the `Analyses used` bullet point, as I found this a bit redundant. Details on analyses included in the notebook could instead be included as a numbered list in the `Description` section (like most notebooks already do).
3. In addition to the above, moved the `Products used` and `Special requirements` bullet points right up to the top next to the `Compatible with...` statement. That way all the important stuff is at the top.
4. Updated the tags to reflect the guide here (https://github.com/GeoscienceAustralia/dea-notebooks/wiki/List-of-tags)

Feel free to edit this branch directly or comment with anything you think should be updated! 

If we settle on any changes, I'm happy to go through the new notebooks and update them to match the revisions before we swap `develop` to `master`.

### DIFF
--- a/DEA_notebooks_template.ipynb
+++ b/DEA_notebooks_template.ipynb
@@ -6,12 +6,18 @@
    "source": [
     "### General advice (delete this cell before submitting for review)\n",
     "> - When choosing a location for your analysis, **select an area that has data on both the `NCI` and `DEA Sandbox`** to allow your code to be run on both environments. \n",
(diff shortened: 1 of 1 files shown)


erialC-P:
Awesome. I’ll try and give it a look tomorrow. ☺

BexDunn:
I'm very happy about the compactifying!

robbibt:
I think I'll merge this in for now so the template is updated for new notebooks, but we can easily make more edits after the Thursday working group meeting!

PR #413, opened 2019-10-10
Notebook demonstrating cloud masking

uchchwhash:
### Proposed changes
- Add notebook showing how to create and apply masks

### DIFF
--- a/Scripts/dea_datahandling.py
+++ b/Scripts/dea_datahandling.py
@@ -23,6 +23,7 @@ Functions included:
     mostcommon_utm
     download_unzip
     wofs_fuser
+    dilate
 
 Last modified: October 2019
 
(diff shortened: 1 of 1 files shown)


BexDunn:
@uchchwhash why does fmask automatically mask water? I think that's not desirable behaviour for many of the science analyses.

uchchwhash:
@BexDunn it does not automatically mask water though. It classifies the water pixels, and it is upto you whether you want to keep it or not. You could set your mask to be valid or water or snow... if your application mandated it.

Do you have any changes to the notebook in mind about this?

uchchwhash:
@robbibt in case you are reviewing this, I updated the notebook after talking to @BexDunn, she had a couple of very helpful suggestions that have since been incorporated into the notebook.

robbibt:
@uchchwhash Yep, I've just checked out version 47cc65e, is that OK? The combination of the three flags change is something I was going to comment about too, so that's a great addition

robbibt:
@uchchwhash I just pushed an update to the notebook. The main change is some extra background up the top, and I moved the `dilate` function to `dea_datahandling` so that it's re-usable. Can you check/revise the documentation for that function to make sure I got everything right? 

Other than that, happy to accept this!

uchchwhash:
Looks ok to me! :+1:

PR #419, opened 2019-10-17
Add tidal phase calculation to tidal_tag function

robbibt:
Modified tidal tag function to add tidal phase calculation.

This allows the user to estimate whether each satellite image was acquired during ebb or flow tides using `ebb_flow=True`. This information is added as a new variable in the dataset.


robbibt:
@erialC-P Fixed! Can you re-do the review (it dismisses it once you make any further changes)

erialC-P:
No worries ☺

PR #423, opened 2019-10-23
notebook now works on Sandbox, edited to reflect this

cbur24:
Datacube stats has been installed on the Sandbox, so this notebook has been updated to reflect that. Also checked for sphinx known-issues and edited markdown.


robbibt:
Updated the style to match the beginners notebooks (we need to update the template to account for this too)

PR #424, opened 2019-10-23
Minor style updates and fixes for readthedocs issues

robbibt:
### Proposed changes
Made some minor style updates for compatability with the new beginners guide notebooks.

Also fixed up some issues that are affecting correct rendering on the readthedocs page (headings order, spaces between bullet points and text, multiple formatting types).

Finally, re-wrote some of the instructional code so that it makes sense even if the text is left in the notebook rather than removed (as happens reasonably regularly.


robbibt:
Note that the DEA logo doesn't render in the template until it's moved to a subdirectory due to the relative file link

PR #439, opened 2019-11-13
Changes to LinearModels-NDVI recession

pkilgour:
Removed global variable (at Neil's request) and added code to export images created with relevant metadata.


BexDunn:
Hi Penny! I think you have pushed this to master instead of to your branch. This PR (pull request) contains your work area, and I think you'd like this to be on your branch but not on master? 

What you probably want to do is create a new temp branch, checkout your updated notebook from Penstemp, then push ONLY that notebook to master from penstemp2 (or what the new branch is named)

Regards,
Bex

PR #447, opened 2019-12-02
Add "Show image on a web map" example notebook

Kirill888:
### Proposed changes

Add notebook that demonstrates interfacing with `ipyleaflet` to show DEA loaded data on a web map.

Only tested on Sandbox, should probably work on NCI, maybe with different product name. Does include configuring for S3 access cell, that will have to **NOT** run on NCI.


Kirill888:
Alright, `show_datasets` accepts `dst=<ipyleaflet.Map>`  so I'll use that and create Map in the notebook itself.

I will replace sharing instructions with a simpler approach from lab that now works. I do think there is still the problem of possibly getting widget state embedded twice, I will test that.

I think I will also add `isSandbox()` and `isNCI()` methods and will wrap rio config in `if isSandbox():` guard.

Kirill888:
I didn't end up doing `isNCI|isSandbox` functions, just checked presence of `AWS` specific env.variables instead. It should now just run on NCI, but I'm not sure how to test.
Also ended up not using `show_datasets` at all, instead done it all in the notebook.

PR #450, opened 2019-12-04
Updated dea-notebooks readme

robbibt:
Work in progress update to the DEA notebooks readme in preparation for switch from `develop` to `master`


robbibt:
@caitlinadams 

- Is it worth linking to the ODC and xarray repos/docs when they're mentioned in the intro paragraph?
Done

- Once this is merged, make sure you check that the internal links work -- since these can't be checked right now.
Will do

- Is it worth expanding the "Working branches" to include both individuals' branches and branches that are dedicated to specific notebooks/features? For example, when someone wants to contribute a new notebook, it's not a good idea to merge their entire individual branch into master. This might overcomplicate the explanation though, so I'm happy to leave it up to you.
I think this should be addressed in the wiki Git guides. It's already in the NCI guide, but needs to be included in the Sandbox guide
 
- As mentioned, it's probably worth revisiting the existing git page so its less Sandbox specific -- or possibly add a page to list ways to build the environments that can run these notebooks. I think this one might be a task for another time, so feel free to ignore this one for now.
- Yep, agree that this needs to be done, but we can work on this in the wiki after the readme is updated 

- Under the publishing section, is it worth mentioning that people can use the "Draft PR" feature for notebooks they'd like additional help with?
Have added a section on draft PRs

PR #454, opened 2019-12-18
Add missing Imagery_on_web_map.ipynb notebook to index

robbibt:
Kirill's notebook was missing from the index and not rendered in the docs
Imagery_on_web_map.ipynb


caitlinadams:
Good catch -- sorry I forgot to do that one!

PR #456, opened 2019-12-20
Minor fixes after ODC Masterclass workshop

caitlinadams:
- remove link to explorer since S1 product doesn't currently appear there
- fix link to internal notebook that covers products and measurements


caitlinadams:
These have been resolved by the notebooks overhaul

PR #462, opened 2020-01-08
Machine learning with ODC

s-m-t-c:
### Proposed changes
This is an example notebook to demonstrate how to use the `dea_classificationtools` for machine learning landcover classification. It attempts to be a simple example to get started, open to feedback regarding complexity etc.

### Checklist


s-m-t-c:
@alexgleith here you go, bit rough around the edges but it works.

caitlinadams:
I'm about to start reviewing this

alexgleith:
The pydotplus package is missing, so we should get that installed in the Sandbox Docker image.

I think having a RGB plot would make the results of the classification easy to evaluate, I used this:
```
sys.path.append('../DEA_sandbox/Tutorial')

from utils.dea_plotting import rgb

rgb(geomedian_data, bands=['swir1', 'nir', 'red'], index=0, size=10)
```

But there may be a neater way to do it. Using `swir1` in the false colour image makes sense as it's the important band for the classification, and you can kind of see how the agriculture stands out in this view.

It's a really neat application. One potential improvement is a bit more context about what is going on and the learning experience around it. Also exploring what the training data looks like? I see it's simply an extent and a column with a class, so maybe that could be described. Also, the test data isn't used (that I noticed), so maybe actually doing the evaluation of the test data would be good. Otherwise we could just use all the training data to train the model, right?

I'll stop there as Caitlin has provided a lot of comments I see!

Thanks :-)

robbibt:
This is a super cool notebook, such a good bit of functionality to have on DEA Notebooks! 
@caitlinadams and @alexgleith have pretty much captured all of my major comments, I think I'll hold off until you've made the recommended changes above, then check out the notebook and make my suggestions/updates directly as a commit. Can you tag me/let me know when there's an update?

s-m-t-c:
This is excellent feedback! Thank you. I will get started on the changes now and add some responses and further questions below. In particular it's good to get feedback on the level of detail / assumed knowledge around machine learning as I wasn't quite sure how much to include.

Apologies for some of the silly mistakes, tried to get something functional out the door quickly and missed some basic things. Maybe not a great idea to submit pull requests at 11 pm.

@caitlinadams 

- The resource I linked is actually available for free on the authors [website ](https://web.stanford.edu/~hastie/pub.htm) but I was torn between linking to that and the DOI of the publication (as persistent links were suggested for the refresh). Maybe I should link both.

- Regarding the `pydotplus `package, I was using it as it makes the tree diagram nice and pretty however there is a builtin one via `sklearn` that doesn't have colours. Maybe I should switch to this as it makes the code a bit simpler and doesn't requite additional packages.

-  I do find that the prediction results are hard to interpret in a notebook and I do most of that in GIS software. Maybe a display map that plots the predicted extent over a basemap would be a good in-between.

@alexgleith 

 - Good point about the training data, will add further detail. As Caitlin suggested maybe a display map showing the shapefile over a basemap would be useful.

alexgleith:
I think viewing the shapefile over a basemap would be great. Also labelling classes somehow would be cool.

I loaded it in QGIS, and it all makes sense. But I think it's worth putting the extra work in to make these notebooks really shiny!

image

s-m-t-c:
I have responded to the suggestions except for those regarding the final plot. I had some troubles with the leaflet method for plotting imagery over a basemap, will sort that out soon as I think getting it looking nice is key for users understanding the process. Also, sorting out issue https://github.com/GeoscienceAustralia/dea-notebooks/issues/477 would help improve the initial displaymap

PR #482, opened 2020-01-17
Geomedian filmstrip plots of change through time notebook

robbibt:
### Proposed changes
Added a new notebook (based on an old notebook) that allows a user to select an area from an interactive map, and generate a geomedian filmstrip plot that highlights change through time.

This uses dask to improve performance; this may work on the NCI but for now I've marked it as a Sandbox-only notebook.

For coastal applications, users can specify a specific portion of the tidal range to generate geomedians from (e.g. low tide, high tide).

filmstrip_example


caitlinadams:
This looks like a super cool notebook! I was wondering whether it belongs more in Frequently_used_code though? While you've got a good background section, there isn't much analysis/commentary about how the film-strip supports a real-world decision/action. Perhaps adding some of this context would round it out as a Real_world_examples notebook? It might also benefit from a title that ties it more to the real world context, rather than the specific analysis, but that doesn't currently hold true for all real world notebooks either (so, up to you).

robbibt:
@caitlinadams It's a tricky one! I think I see the notebook fitting in `Real world examples` because it's designed as a functional tool for obtaining real-world insights from satellite data... whereas `Frequently used notebooks` provide minimum viable code for performing common tasks. This tool is something that's come up frequently from various stakeholders as something they would find useful, so it's quite similar to Katie's inundation notebook in that way. The code in this one is also all locked up inside the "app" function to make it easier to use, while `Frequently used notebooks` generally walk through the different steps in a bit more detail, maybe? 

To address this a bit I have now added some extra interpretation markdown at the end of the notebook which provides some extra info about the kinds of insights the filmstrip plots can help produce, and also the "Next steps" section which we have in the other `Real world examples notebooks`.

robbibt:
The title could definitely be improved... even a simple change to something like `Analysing change through time using filmstrip plots" would help shift the focus onto being able to study change vs. the specific technique

robbibt:
@BexDunn I've just added this to make the load and status a bit clearer:

image

image

BexDunn:
> @BexDunn I've just added this to make the load and status a bit clearer:
> 
> image
> 
> image

Looks good to me, do you want me to squash and merge or do you want to wait for other reviews/ change the title/ etc?

robbibt:
@BexDunn Sean wanted to have a look at it first, so maybe mark as "accepted" so it knows I addressed the requested changes, but leave it un-merged?

Issue #485, opened 2020-01-23
Add nbar(t) contiguity masking to load_ard


cbur24:
When loading GA Landsat Collection 3 data, the current iteration of `dea_datahandling.load_ard ` masks invalid pixels using only `fmask`.  Including masking by the `oa_nbar(t)_contiguity` band would further reduce the number of noisy pixels included in the returned dataset.  Agnostic about whether this should be by default or added as an optional flag.

robbibt:
Fixed by #487

Issue #433, opened 2019-11-06
Clean up confusing dask API in load_ard


robbibt:
Currently, `load_ard` uses `dask_chunks={'time': 1}` to force data to lazy load before scenes are filtered out based on the proportion of valid/non cloudy data in each timestep. To override this, users can set the function's `dask_chunks` parameter. However, if a custom `dask_chunks` is also passed in via a query, this has no effect on the analysis. 

Additionally, even if custom `dask_chunks` is passed in via a query, unless the `load_ard` kward "lazy_load' is set to True, data will still compute.

This combination of dask-related function kwargs and query kwargs is confusing (e.g. below) and needs to be streamlined.

![image](https://user-images.githubusercontent.com/17680388/68261108-2c33b480-0093-11ea-91f9-de508b50deb9.png)

robbibt:
Fixed by #487

Issue #360, opened 2019-09-19
Make masked dtype configurable for `load_ard`


robbibt:
Add the ability to customise the dtype of the output data from `load_ard` when masking cloud/invalid data to provide more flexibility for advance users

> should we make this configurable? maybe like `masked_dtype='float32'` with ability to overwrite with `float64|16` by users who need more control over memory/precision trade of.
> 
> _Originally posted by @Kirill888 in https://github.com/_render_node/MDExOlB1bGxSZXF1ZXN0MzE5MTIzMjY0/pull_requests/unread_timeline_

robbibt:
Fixed by #487

PR #487, opened 2020-01-24
Revise load_ard to address #360, #433, #485

robbibt:
This PR is an update to the `load_ard` function, addressing several issues:

#360 Make masked dtype configurable for `load_ard`:
- Addressed by adding a custom parameter `mask_dtype` that allows the user to specify a specific dtype when masking data. This can be used to save memory when loading large datasets.

#433 Clean up confusing dask API in load_ard
- Addressed by removing the `dask_chunks` and `lazy_load` params, and replacing them with code that detects whether the user has supplied `dask_chunks` in the datacube query, then lazily loads data accordingly

#485 Add nbar(t) contiguity masking to load_ard
- Have changed the older `mask_invalid_data` parameter to `mask_contiguity` which sets pixels to NaN if they are missing a value in any band
- This works nicely for Landsat, but has some issues due to the current 20m resolution implementation of `nbar(t)_contiguity` in Sentinel-2, which doesn't appropriately mask out nodata in the Sentinel-2 10m bands
- I've currently added in a`mask_invalid_data` step too to account for this... could go back to allowing the user to specifically choose to conduct this step if they want (rather than including it in the `mask_contiguity` step...)

I've also updated any notebooks that use `load_ard`. I've tried to minimise breaking changes, but unfortunately if you used the previous `lazy_load` parameter you'll need to update your code (e.g. replace it by supplying `dask_chunks` only).


robbibt:
@cbur24 This adds in the continuity stuff, and fixes up the dask/lazy loading API so you essentially use it just like dc.load by specifying `dask_chunks` in the query. Do you have a dask analysis workflow you could test this out on?

robbibt:
After talking to Imam about an issue with contiguity on 10m Sentinel-2 data, I'm going to make an additional update to this today to restore the `mask_invalid_data` param I replaced with `mask_contiguity`. Essentially, the contiguity band doesn't correctly screen out all nodata values when applied to 10m Sentinel-2 data (even though it theoretically should). We therefore still need the `mask_invalid_data` option to catch all stray -999s.

robbibt:
@cbur24 Just added the `mask_invalid_data` back in... any chance you could re-test it works for you then accept and merge using `squash and merge`?

Issue #477, opened 2020-01-13
Allow map_shapefile to plot colours based on attribute


robbibt:
Currently, the `map_shapefile` function (https://github.com/GeoscienceAustralia/dea-notebooks/blob/master/Scripts/dea_plotting.py#L330) plots each feature in a shapefile with an incrementing colour. 

It would be far more useful if this function plotted features based on values from a field in the shapefile attribute table.

s-m-t-c:
The next steps are to develop some logic regarding the treatment of categorical and continuous variables and use some of the ideas from this leaflet [example](https://leafletjs.com/examples/choropleth/). This will be inserted near line 392 as an option enabled when a string containing the attribute column name we wish to colour the features by is passed into the function.

Need to spend some more time thinking about how to bin and organise the colours for continuous values. This will probably live in a `get_colour` function similar to that shown in the linked example.

robbibt:
@s-m-t-c The above sounds great to me. For the simplest continuous numbers case, I imagine we want to generate a unique, linearly spaced colour stretched between the lowest and the highest values? So, kind of linearly sampling a colour map based on the values in the attribute table? More complex binning (e.g. assigning a specific colour to appear at a specific value) is probably possible but a fair bit more complicated

robbibt:
@s-m-t-c Categorical mightn't be that hard to implement: if there's e.g. 4 classes/unique values in the attribute column, simply sample the cmap four times, evenly spaced?

s-m-t-c:
@gabzgit for your reference

PR #499, opened 2020-02-05
modify map_shapefile to colour by attribute value

Gabzgit:
### Proposed changes
Adds the gdf as a choropleth onto the ipyleaflet. Requires a column name to colour by. 

- If remap is false, it will colour based on the values in the specified column. This is better for continuous variables.

- If remap is true (default) the values will be changed for better visualization. This is better for categorical data. For example, if original values are 111, 112 and 215, on a colour ramp 111 and 112 will be very similar. So remap to 0, 1 and 2.

### Closes issues (optional)
Helps Issue #477

### DIFF
--- a/Scripts/dea_plotting.py
+++ b/Scripts/dea_plotting.py
@@ -35,11 +35,13 @@ import ipywidgets
 import matplotlib as mpl
 from pyproj import Proj, transform
 from IPython.display import display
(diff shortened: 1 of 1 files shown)


s-m-t-c:
@robbibt if you can check this function hasn't broken any of the previous capabilities of the one you wrote + general feedback that would be great.

robbibt:
Superseded by PR #506

PR #500, opened 2020-02-05
Add notebook from SAR workshop

whatnick:
### Proposed changes
Add SARA based SLC download reusable code


robbibt:
Hi @whatnick , I'm just going through this now. Some early comments/questions... will post extra ones later on: 

1) What does` !export DISPLAY=0:0` bit do when you're installing the custom packages? It comes up again near the end; some extra info to help the user know what's going on here might be good
2) How slow is "slow"? And how big are the files that are being downloaded? If it's huge/long, we might want to think about reducing the study area or the temporal range so that the notebook runs quickly (and add some extra info to the markdown giving users a very rough estimate of time/total amount of data in this example

whatnick:
> Hi @whatnick , I'm just going through this now. Some early comments/questions... will post extra ones later on:
> 
> 1. What does` !export DISPLAY=0:0` bit do when you're installing the custom packages? It comes up again near the end; some extra info to help the user know what's going on here might be good

It allows install of PyRAT which requires PyQT5 in headless mode by supplying a fake display. See discussion on this in PyRAT github here: https://github.com/birgander2/PyRAT/issues/18

> 2. How slow is "slow"? And how big are the files that are being downloaded? If it's huge/long, we might want to think about reducing the study area or the temporal range so that the notebook runs quickly (and add some extra info to the markdown giving users a very rough estimate of time/total amount of data in this example

Slow depends on network available on NCI. Actually on Sandbox this is not very slow. Order of 10 minutes or so.

robbibt:
OK, have submitted some direct changes on the forked repo. 

 @whatnick It's a great notebook - there was just a few more minor things that would improve the notebook and make it easier to use for new users who might not be familiar with SAR:

1. In the "background" section, it would be great to have a simple explanation of what SLC is, including a very brief (1 sentence?) description of Sentinel-1/radar

2. Similarly, a brief description of what SARA is in "background" would be great

3. Without having any domain knowledge of Sentinel-1 data or SLC, I'm unsure of what the image snippet plots and the histogram plots in the notebook are meant to show... the notebook says "as actual fire based variability is noted", but it would be great to explain this a little more next to the plots to users know what they are meant to be looking for. 

(Not essential, but it would be nice if the image snippet plots and the histogram plots were displayed in the same orientation/order to make comparisons easier - right now, one is 3 x 2 and the other 2 x 3)

4. Finally, the PyRAT bit at the end could also do with some additional explanation to let users understand what that part of the analysis is aimed at achieving, and why it is useful. I downloaded and inspected the two images it produced, but without Sentinel-1 knowledge I didn't know what I would use them for.

Happy to merge after those changes! :)

PR #506, opened 2020-02-10
Modify map_shapefile to colour by attribute value 

robbibt:
### Proposed changes
I've just created a new PR that follows on from #499 as the fork it was based on doesn't exist any more. This includes all the main changes implemented by @Gabzgit , plus a few extras on top:

**gabzgit changes:**

- Adds the gdf as a choropleth onto the ipyleaflet. Requires a column name to colour by.
- If remap is false, it will colour based on the values in the specified column. This is better for continuous variables.
- If remap is true (default) the values will be changed for better visualization. This is better for categorical data. For example, if original values are 111, 112 and 215, on a colour ramp 111 and 112 will be very similar. So remap to 0, 1 and 2.

**robbibt changes:**

- Polyline shapefiles plot as black only in `map_shapefile` due to the Choropleth function only controlling fill colour. To address this, I've used a bit of a hack by extracting the raw geojson using `choropleth.data`, copying fill colours to line colours, and then plotting this directly for line features.
- Have renamed `remap=True` to `continuous=False` to make it it a bit more intuitive to choose the right option for your data
- Added a `**style_kwargs` kwarg so that users can pass in custom style params
- Based `hover_col` off the field selected by `attribute` for simplicity

I've updated all the notebooks based on `map_shapefile` so they work with the new function. Example of it in action in the machine learning notebook (@s-m-t-c ):

image

### Closes issues (optional)
- Closes Issue #477


robbibt:
@Gabzgit Good idea, will add this as an option

robbibt:
@Gabzgit I've updated it so the default `hover_col` value True will use the `attribute` field, but the user can pass in a custom string giving a different attribute field if they want to

PR #516, opened 2020-02-13
Add dNBR notebook

Gabzgit:
### Proposed changes
Add Delta Normalized Burn Ratio notebook into Real_world_examples 


cbur24:
Really nice work, Gabrielle! This will be a really valuable notebook in the repo.

I've made a few minor changes including:
- fixing typos in the markdown, and highlighting a few important notes for the user
- Using `dea_bandindices.calculate_indices` function for the post-fire NBR calculation to be consistent with the pre-fire index calculation.
- using `round` on the printed burnt areas for neatness

Couple of other things to consider:
1. The facet plot that's supposed to help for adjusting the threshold is confusing and could use more markdown and/or titles on each plot.  And the facet-plot has 6 axes but only 5 plots. Is it missing a true-colour image?   The `rgb` function can now accept an `ax` argument so you can add a nice true-colour plot to the facet-grid quite easily.

2. You calculate the post-fire NBR on a `.median()` of the input time-series, whereas for the pre-fire NBR you first calculate NBR on each image and then calculate the median. I suggest being consistent with the approaches.  Unless implementing `geomedians`, probably best to calculate NBR on each image and then calculate the median NBR for the post-fire time-series.

Nearly there!

robbibt:
This is a super nice notebook - I've made some markdown updates and some small tweaks to the code:
1. Use f-strings for string formatting/rounding
2. Remove the axis labels from `.imshow()` plots
3. Use `xarray` stats rather than `numpy` stats where possible

I really like the detailed intro and the six panel comparison panel at the end. Great work - if no-one else has anything to suggest, I think this is ready to be accepted/merged!

PR #518, opened 2020-02-17
Fix image segmentation notebook

Kirill888:
### Proposed changes

Fix error in `Image_segmentation.ipynb` one of the output files was named with `.tif` but was populated by `rsgislib` with KEA format, `rsgislib` would then fail to open the file it produced to add statistics and overviews to it. Changing file name to end on `.kea` fixes the problem. 

This was tested in the sandbox using custom docker with `rsgislib` installed, this docker is not yet available to sandbox users.

### DIFF
--- a/Frequently_used_code/Image_segmentation.ipynb
+++ b/Frequently_used_code/Image_segmentation.ipynb
@@ -357,8 +357,8 @@
     "# Name of the segmented .kea file that will be output \n",
     "segmented_kea_file = 'meanNDVI_segmented.kea'\n",
     "\n",
-    "# Name of the segmented GeoTIFF attributed with the zonal mean of input file\n",
-    "segments_zonal_mean = 'segments_zonal_mean_shepherdSeg.tif'\n"
+    "# Name of the segmented .kea file attributed with the zonal mean of input file\n",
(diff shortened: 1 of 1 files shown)


robbibt:
I get the following error on the NCI (`module load dea/20191127`, can't currently test on `module load dea` as I need `dea/20191127` for some other analysis):

```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-65266bf9980b> in <module>
     10 from datacube.helpers import write_geotiff
     11 from skimage.segmentation import quickshift
---> 12 from rsgislib.segmentation import segutils
     13 
     14 sys.path.append('../Scripts')

/g/data/v10/public/modules/dea-env/20191127/lib/python3.6/site-packages/rsgislib/segmentation/__init__.py in <module>
     26 
     27 # import the C++ extension into this level
---> 28 from ._segmentation import *
     29 #import warnings
     30 #

ImportError: libCGAL.so.13: cannot open shared object file: No such file or directory
```

Kirill888:
@robbibt yep broken conda install, missing or wrong version of one `libCGAL` used by `rsgislib` in that env.

robbibt:
Just tested on `module load dea`, same issue there too. Ping @santoshamohan

Issue #513, opened 2020-02-13
Create `xr_geomedian` notebook for `frequently_used_code` folder


cbur24:
@Kirill888 's `xr_geomedian` [notebook](https://github.com/opendatacube/odc-tools/blob/master/notebooks/geomedian-example.ipynb) would make a nice addition to the `frequently_used_code` folder of dea-notebooks.  The current [Geomedian_composites](https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Frequently_used_code/Geomedian_composites.ipynb) notebook is very computationally expensive, and thus limits the analysis extent that users can apply it too.

robbibt:
Closed by #521

PR #521, opened 2020-02-17
Refactoring the Geomedian Composites notebook #513 

cbur24:
### Proposed changes
I've re-written the `Geomedian_Composites` notebook to work with the `ODC.algo` function `xr_geomedian`. This allows for running the geomedian code in parallel using a dask cluster. Users should now be able to run geomedian code faster, and over bigger spatio-temporal extents, within the notebook environment.

### Closes issues (optional)
- Closes Issue #513 


cbur24:
There is an issue here with using `dask` that I could use some guidance around.  Setting up `dask` on the Sandbox is different from setting it up on the NCI:

1. First of all, this line is not necessary on the NCI
```
dask.config.set({"distributed.dashboard.link":
#                  os.environ.get('JUPYTERHUB_SERVICE_PREFIX', '/')+"proxy/{port}/status"})
```
2. The way you retrieve the `cpu` and `memory` allocations is different between the NCI and Sandbox, on the NCI the `os.environ.get` methods don't work. Instead, something like this is more appropriate:

```
from psutil import virtual_memory, cpu_count

cpu_limit = cpu_count()
cpu_limit = int(cpu_limit) if cpu_limit > 0 else 4
mem_limit = virtual_memory()
mem_limit = mem_limit.total
mem_limit = mem_limit if mem_limit > 0 else parse_bytes('8Gb')
```

3. odc.alog is not installed by default on the NCI (easily solvable - I think).

These issues could be solved by having the user provide a flag (e.g. `sandbox = True`) for whether they're on the NCI or Sandbox, and then the correct `dask cluster` code would be run. But this solution seems messy and the dask code is already a little intimidating and I don't want the notebook to be off-putting to users by having some big chunk of dask code at the beginning.

Any thoughts or advice would be appreciated.

Kirill888:
@cbur24 sorry I need to sort out docker image for sandbox before I can look at this

cbur24:
@robbibt has directed me to this code snippet that should help with automatically identifying the environment the notebook is running in:

`if 'AWS_ACCESS_KEY_ID' in os.environ:`

I will test with this and see how unwieldy the code gets

cbur24:
This automatic selection of cpus and memory for the dask cluster is now working on the Sandbox and NCI.  `odc.algo` is not installed so this notebook doesn't yet work on the NCI

Kirill888:
Started looking into this, looks good, some performance issues I came across stem from use of `load_ard` function. For example when displaying rgb preview `load_ard` loads all the data for a product referenced by that timeslice, this is due to use of pixel stacking inside `load_ard`.

In no particular order

- Resolution is set 30m while input is 10m native
- rgb preview is slow because of `load_ard` issue (I talked to @robbibt about it and will create separate issue for it)
- Worthwhile adding `datacube.utils.rio.configure_s3_access` call when constructing dask `client` (sandbox only), gives some IO improvement
- There is now `mem_safety_margin` option in `start_local_dask` which can be used instead of computing `mem_limit` in the notebooks, it does basically the same logic. Not sure if available on NCI yet.
- I think this notebook needs more discussion about `dask_chunks` parameter, maybe `500` is a bit too small. The trade-off roughly like this:
   - Larger values -- more efficient IO, fewer passes over data, BUT more likely to run out of Memory
   - Smaller values -- more robust to out of memory problems BUT slower, because loading 1 whole image once is faster than loading 1/10 of an image 10 times.

cbur24:
The notebook now implements the 'dea_dask.py' cluster creation function.  Included with this pull request is the removal of the following code snippet from `create_local_dask_cluster`:

```
client = locals().get('client', None)
        if client is not None:
            client.close()
            del client
```
This code is only useful when re-creating the cluster without losing the state in the notebook;
and `locals()` does not work inside a function @robbibt

BexDunn:
how are we doing on this one? Are all the moving pieces assembled with 0dc1.8 yet?

cbur24:
> how are we doing on this one? Are all the moving pieces assembled with 0dc1.8 yet?

This PR needs to be updated as DE Africa repo now has a helpful wrapper function for running `xr_geomedian`.  I can do this but don't know exactly when I'll have time, so feel free to jump in and refactor if you'd like.
- [wrapper function](https://github.com/digitalearthafrica/deafrica-sandbox-notebooks/blob/595f9634371f9046ec294987495fc9acb4d08a31/Scripts/deafrica_datahandling.py#L838)
- [Notebook implementing wrapper](https://github.com/digitalearthafrica/deafrica-sandbox-notebooks/blob/master/Frequently_used_code/Generating_geomedian_composites.ipynb)

Kirill888:
@BexDunn @cbur24 
I'm thinking of changing implementation of the Dask Geomedian to NOT require stacking all bands together, as I found it to be a surprisingly expensive step for Dask (at least for network-distributed version, maybe not so bad locally). Also I learned few things about low-level Dask since then and I think there is a better approach to this. From the API perspective this means that the new implementation will operate on Datasets and not on DataArrays. I'll probably still keep DataArray version around though, for situations when data is natively that way.

cbur24:
@Kirill888 
Is this something likely to be implemented in the short term?  If not, then maybe its still worth finalizing this PR so there is at least a dask-enabled workflow for creating geomedians on the repo, shouldn't be too hard to update the notebook/function if it switches over to accepting a dataset rather than an array.

Kirill888:
Define short term 😄? I think there is no point waiting, this was more of a "heads up" comment from me.

cbur24:
Too many conflict in this branch so closing this PR and will create a new PR

Issue #526, opened 2020-02-19
Make `load_ard` with multi-products more "dask friendly"


Kirill888:
`load_ard` function defined here:

https://github.com/GeoscienceAustralia/dea-notebooks/blob/4891b9167966ede8dca697b8c4d79acb493071d9/Scripts/dea_datahandling.py#L50-L60

supports loading several "structurally similar" products as if it was one product. Current approach is to perform `dc.load` for every product separately then stack results along the time dimension. This approach, while logical and simple, creates some performance issues, particularly when producing dask arrays (i.e. when `dask_chunks={..}` is supplied). But even for non-dask load this approach leads to significantly higher peak memory usage than alternative.

### Proposed Approach

Use `.find_datasets` across all products first, building up a list of datasets, named `dss` (for example), then use `dc.load(datasets=dss, ...)` to load datasets across all products in one go. This is also a more correct approach with respect to "pixel fusing", in the situations when two different products have observations on the same day (not something that happens with current products).

robbibt:
As discussed, I think this makes a lot of sense and it'd be great to have `load_ard` work well for complex dask workflows. 

Here's a working proof-of-concept using the dataset loading method which seems to work, and keeps intact the three main features of `load_ard`:

* Combining multiple products (e.g. ` ['ga_ls5t_ard_3', 'ga_ls7e_ard_3', 'ga_ls8c_ard_3']`)
* Masks data by pixel quality etc
* Filters data to drop cloudy scenes

```
# Load ard params
products = ['ga_ls5t_ard_3', 'ga_ls7e_ard_3', 'ga_ls8c_ard_3']
min_gooddata = 0.9
ls7_slc_off = False
dcload_kwargs = {
    'x': (153.38, 153.47),
    'y': (-28.83, -28.92),
    'time': ('2018-01', '2018-12'),
    'measurements': ['nbart_red', 'nbart_green', 'nbart_blue', 'fmask'],
    'output_crs': 'EPSG:3577',
    'resolution': (-30, 30),
    'group_by': 'solar_day',
    'dask_chunks': {}
}


# Extract datasets for each product using subset of dcload_kwargs
dataset_list = []
datasets_query = {k:v for k,v in dcload_kwargs.items() 
                  if k in ['time', 'x', 'y']}

# Get list of datasets for each product
for product in products:    

    # Obtain list of datasets for product
    datasets = dc.find_datasets(product=product, **datasets_query)    

[...]

PR #527, opened 2020-02-20
Refactor `load_ard` to be more "dask friendly"

robbibt:
### Background
#526 discusses issues with the current implementation of `load_ard` that can cause performance issues in combination with `dask` due to loading individual products then stacking them with `xr.concatenate` (potentially running into problems with mis-aligned chunks and peak memory issues).

### Proposed changes
This PR aims to re-factor `load_ard` using the following workflow:

1. Uses  `dc.find_datasets()` across all products first, building up a simple list of datasets named `dataset_list`
2. Uses `dc.load(datasets=dataset_list, ...)` to load datasets across all products in one go
3. Masks data without converting to floats until the final step using tools from `odc.algo`, avoiding issues with high peak memory usage

This change has greatly simplified the function code, and has also enabled some potentially exciting new features, including the new `filter_func` parameter. This lets you pass in a custom function which filters out individual datasets prior to them even being loaded using `dc.load`. For example, to **load only summer months** from multiple years:

```
query = {
    'x': (153.38, 153.47),
    'y': (-28.83, -28.92),
    'time': ('2015', '2018'),
    'measurements': ['nbart_red', 'nbart_green', 'nbart_blue'],
    'output_crs': 'EPSG:3577',
    'resolution': (-30, 30),
    'group_by': 'solar_day',
    'dask_chunks': {}
}

def month_filter(dataset, months=[12, 1, 2]):    
    # Return True if dataset month is in list of months
    return dataset.time.begin.month in months

ds = load_ard(dc=dc,
              products=['ga_ls5t_ard_3', 'ga_ls7e_ard_3', 'ga_ls8c_ard_3'],
              filter_func=month_filter,
              **query)
```

### Closes issues (optional)
- Closes Issue #526


cbur24:
I ran this through the `load_ard` notebook without fault.  I also ran it through the `geomedian_refactor` notebook and it resulted in a smallish (~10%) improvement in the total wall-time. The `filter_func` parameter worked as expected during loading of a sentinel 2 dataset.

Kirill888:
@cbur24 thanks was this on NCI or sandbox?

I still need to fix how parameters for load vs query are split, currently code only allows querying time,x,y which is not right.

cbur24:
I ran the `geomedian-load_ard` test on the Sandbox, using 14Gb of RAM and 2 cores (1-worker). 

```
query = {
    'x': (153.35, 153.50),
    'y': (-28.85, -29.00),
    'time': ('2017'),
    'measurements': ['nbart_green',
                     'nbart_red',
                     'nbart_blue'],
    'output_crs': 'EPSG:3577',
    'resolution': (-10, 10),
    'group_by': 'solar_day'
}

ds = load_ard(dc=dc, 
              products=['s2a_ard_granule', 's2b_ard_granule'],
              dask_chunks={'x': 1000, 'y':1000},
              **query)
```

Kirill888:
Some more potential issues with `load_ard` when not constraining `measurements` and requesting "landsat" different products might have different set of bands.

robbibt:
Just tested out the latest update using the example in the `Using_load_ard.ipynb` notebook and got the following error:

```
# Load available data from all three Landsat satellites
ds = load_ard(dc=dc_landsat,
              products=['ga_ls5t_ard_3', 'ga_ls7e_ard_3', 'ga_ls8c_ard_3'],
              x=(153.38, 153.47),
              y=(-28.83, -28.92),
              time=('2018-04', '2018-06'),
              measurements=['nbart_green'],
              output_crs='EPSG:3577',
              resolution=(-30, 30),
              group_by='solar_day')

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/usr/local/lib/python3.6/dist-packages/xarray/core/dataset.py in _construct_dataarray(self, name)
   1154         try:
-> 1155             variable = self._variables[name]
   1156         except KeyError:

KeyError: 'fmask'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
<ipython-input-3-7dc9d1d3a6ec> in <module>
      8               output_crs='EPSG:3577',
      9               resolution=(-30, 30),
---> 10               group_by='solar_day')
     11 
     12 # Print output data

~/dev/dea-notebooks/Scripts/dea_datahandling.py in load_ard(dc, products, min_gooddata, fmask_categories, mask_pixel_quality, mask_contiguity, ls7_slc_off, filter_func, **extras)
    305 
    306     # Calculate pixel quality mask
--> 307     pq_mask = odc.algo.fmask_to_bool(ds[fmask_band],
    308                                      categories=fmask_categories)
    309 

/usr/local/lib/python3.6/dist-packages/xarray/core/dataset.py in __getitem__(self, key)
   1243 
   1244         if hashable(key):
-> 1245             return self._construct_dataarray(key)
   1246         else:
   1247             return self._copy_listed(np.asarray(key))

/usr/local/lib/python3.6/dist-packages/xarray/core/dataset.py in _construct_dataarray(self, name)
   1156         except KeyError:
   1157             _, name, variable = _get_virtual_variable(
-> 1158                 self._variables, name, self._level_coords, self.dims
   1159             )
   1160 

/usr/local/lib/python3.6/dist-packages/xarray/core/dataset.py in _get_virtual_variable(variables, key, level_vars, dim_sizes)
    163         ref_var = dim_var.to_index_variable().get_level_variable(ref_name)
    164     else:
--> 165         ref_var = variables[ref_name]
    166 
    167     if var_name is None:

KeyError: 'fmask'
```

cbur24:
Two current issues with load_ard:
1. When providing `measurements=` to `load_ard`, a `keyError` is returned for the cloud and continguity masking steps. This is because both bands require "oa_" at the start of the band name, but the prefix is only added if measurements _isn't_ supplied.

2. The returned datasets are much larger than the specified query. The following query returned a dataset with dimensions: `(time: 16, x: 13486, y: 9417)`

```
        query: {'x'=(153.38, 153.47),
                    'y'=(-28.83, -28.92),
                   'time'=('2018-04', '2018-06')}

```

Kirill888:
Thanks, yeah, x,y,lat,lon needs to go to load as well as query, will fix that. The whole masking band selection needs to change I reckon it's too confusing to get right, I thought that `oa_` prefix is only needed because aliases do not get used when loading all bands, but are used when loading with set names.

Kirill888:
oh and `geopolygon` too

cbur24:
So adding `**query` to the  dc load did the trick on the second issue:
```
ds = dc.load(datasets=dataset_list,
                 dask_chunks={} if dask_chunks is None else dask_chunks,
                 **load_params, **query)
```

Is that the right way to do it?

Kirill888:
Turns out I was using wrong branch for testing, I think I checked in the fix @cbur24 @robbibt

Kirill888:
Well it was too hard to figure out what parts of query `.load(..)` needs to construct proper bounding box, so I gave up on that and instead only extract query parameters for `.find_datasets` and pass everything to `.load` except for `measurements` and `dask_chunks` which get modified by `load_ard` code.

cbur24:
Ran the current iteration through the `Using_load_ard` notebook with no issues.

Kirill888:
There is still a problem, which I think was there before the refactor: `fmask` will be loaded and then dropped even if user didn't ask to filter data and didn't request `fmask` band.

Also currently masking is applied to both data bands and mask bands, which is not what I would expected if I were to request masking and also `fmask` band to be returned.

Also there are now a lot of garbage commits, we should rebase, I won't do it today in case people have uncommitted code, but we need to do it before merging to develop.

robbibt:
Thanks @Kirill888 for all the updates, everything seems to be running nicely on my end. Will test it out on the repo notebooks later this week to make sure everything still works. Some questions:

> There is still a problem, which I think was there before the refactor: fmask will be loaded and then dropped even if user didn't ask to filter data and didn't request fmask band.

Is this a problem if the `fmask` band is loaded and dropped lazily? Or does it add extra overhead?

> Also there are now a lot of garbage commits, we should rebase, I won't do it today in case people have uncommitted code, but we need to do it before merging to develop.

We usually "squash and merge" when merging into `develop`... given that squashes the commit history, is there an advantage to rebasing instead?

One extra question: the `dcload_kwargs` is now renamed `extras`... I think I preferred the previous name as it is a bit more explanatory (the first time people use `load_ard` they often get confused about the fact that it takes `dc.load` parameters directly)... was there a reason for the name change?

Kirill888:
@robbibt 

Re mask probably it's fine, still I think it needs separate handling. I just looked again and see a problem with conversion to float, masks should not be converted to float, it's not what ones want. 

Re: "squash and merge", I'd rather not in this case, I have cleaned up a lot of whitespace, which touches a lot of lines. I'd rather keep that commit as a separate commit.

Re `dcload_kwargs` - well it's no longer correct name, majority of those params go to `find_datasets` and they are not just about load they are about query. These are hard to name. In a lot of python libs it's common to just call those `**kw` or `**kwargs` (kw- key word). How about we just call it `**kwargs`.

Other naming related thing `filter_func`, arguments like that are usually called `pred|predicate`, which is what it is. Example: `https://docs.python.org/3/library/itertools.html?highlight=pred#itertools.dropwhile`

Kirill888:
More or less happy with the current state:

- no longer messing up `fmask` output (do not mask it, do not convert to float)
- will load in native `dtype` if no masking is applied (same as previous behaviour)
- added back `dtype` parameter, used to be `mask_dtype`, this can be `auto` (default), `native` or `float{16,32,64}` (changes to `odc.algo` were needed to support non-float32 outputs)
- `dtype='auto'` means do `float32` if masking do `native` if not.
- `dtype='native'` means you can have cloud masking applied, but still get back pixels in native format (`int16` usually). `nodata` is used to mark clouds in this case.

To test in sandbox need to upgrade `odc.algo`: 
```
pip install -U --user --extra-index-url="https://packages.dea.ga.gov.au/" odc-algo
```

robbibt:
Changes:

* Have re-written some of the code docstring and added some extra comments for clarify, 
* Re-ordered some of the function so it follows a more logical flows (setup > find datasets > load data > filter good data > apply masks > return data).
* Updated the `Using_load_ard` notebook to document some of the new features

Everything seems to be working nicely - I've tested the updated function against all Frequently used code and Real world examples notebooks, and no errors have shown up.

Regarding merging this in, I think we will probably need to wait until `odc.algo` is upgraded on both Sandbox and NCI. @Kirill888 , is the upgrade included in the proposed new image for the Sandbox?

Kirill888:
@robbibt `odc-algo` with fixes is already in our internal packages repo, so it will be pulled next time any of the sandbox images are updated (new or old).

robbibt:
Hmm, so testing this out on the NCI and running into some issues. This code works correctly on the Sandbox:

```
from dea_datahandling import load_ard

lat, lon = -20.58, 117.87

# Load available data from all three Landsat satellites
ds = load_ard(dc=dc,
              products=['ga_ls5t_ard_3', 'ga_ls7e_ard_3', 'ga_ls8c_ard_3'],
              x=(lon - 0.01, lon + 0.01),
              y=(lat - 0.01, lat + 0.01),
              time=('1988', '2018'),
              measurements=['nbart_green'],
              output_crs='EPSG:3577',
              resolution=(-30, 30),
              group_by='solar_day',
              dask_chunks={})
```

But the same code produces the following error on the VDI (after `module load dea`):

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-6-9a1ca26e2bd5> in <module>
     13               resolution=(-30, 30),
     14               group_by='solar_day',
---> 15               dask_chunks={})
     16 

/g/data1a/r78/rt1527/dea-notebooks/Scripts/dea_datahandling.py in load_ard(dc, products, min_gooddata, fmask_categories, mask_pixel_quality, mask_contiguity, ls7_slc_off, predicate, dtype, **kwargs)
    329                  measurements=measurements,
    330                  dask_chunks={} if dask_chunks is None else dask_chunks,
--> 331                  **kwargs)
    332 
    333     ####################

/g/data/v10/public/modules/dea/20191127/lib/python3.6/site-packages/datacube/api/core.py in load(self, product, measurements, output_crs, resolution, resampling, skip_broken_datasets, dask_chunks, like, fuse_func, align, datasets, progress_cbk, **query)
    298 
    299         geobox = output_geobox(like=like, output_crs=output_crs, resolution=resolution, align=align,
--> 300                                grid_spec=self.index.products.get_by_name(product).grid_spec,
    301                                datasets=observations, **query)
    302 

AttributeError: 'NoneType' object has no attribute 'grid_spec'
```

Sandbox datacube.__version__:  '1.7+255.gdb5854af.dirty'
NCI datacube.__version__: '1.7+142.g7f8581cf'

robbibt:
@Kirill888 Hmm, so testing this out on the NCI and running into some issues. This code works correctly on the Sandbox:

```
from dea_datahandling import load_ard

lat, lon = -20.58, 117.87

# Load available data from all three Landsat satellites
ds = load_ard(dc=dc,
              products=['ga_ls5t_ard_3', 'ga_ls7e_ard_3', 'ga_ls8c_ard_3'],
              x=(lon - 0.01, lon + 0.01),
              y=(lat - 0.01, lat + 0.01),
              time=('1988', '2018'),
              measurements=['nbart_green'],
              output_crs='EPSG:3577',
              resolution=(-30, 30),
              group_by='solar_day',
              dask_chunks={})
```

But the same code produces the following error on the VDI (after `module load dea`):

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-6-9a1ca26e2bd5> in <module>
     13               resolution=(-30, 30),
     14               group_by='solar_day',
---> 15               dask_chunks={})
     16 

/g/data1a/r78/rt1527/dea-notebooks/Scripts/dea_datahandling.py in load_ard(dc, products, min_gooddata, fmask_categories, mask_pixel_quality, mask_contiguity, ls7_slc_off, predicate, dtype, **kwargs)
    329                  measurements=measurements,
    330                  dask_chunks={} if dask_chunks is None else dask_chunks,
--> 331                  **kwargs)
    332 
    333     ####################

/g/data/v10/public/modules/dea/20191127/lib/python3.6/site-packages/datacube/api/core.py in load(self, product, measurements, output_crs, resolution, resampling, skip_broken_datasets, dask_chunks, like, fuse_func, align, datasets, progress_cbk, **query)
    298 
    299         geobox = output_geobox(like=like, output_crs=output_crs, resolution=resolution, align=align,
--> 300                                grid_spec=self.index.products.get_by_name(product).grid_spec,
    301                                datasets=observations, **query)
    302 

AttributeError: 'NoneType' object has no attribute 'grid_spec'
```

Sandbox `datacube.__version__`:  '1.7+255.gdb5854af.dirty'
NCI `datacube.__version__`: '1.7+142.g7f8581cf'

Kirill888:
Thats because this commit:

https://github.com/opendatacube/datacube-core/commit/b4d5647dc8e488cb3f5ca7bc94ade8310f6a7239

is not on NCI env yet I guess

PR #537, opened 2020-03-31
Fix possible bug in subpixel_contours

CEKrause:
`crs` is defined in the function a few lines above as a crs object.
Later on it then takes this as a string and tries to feed it into the `{'init:(crs)}` format, but this errors because the crs is not just an EPSG number.


robbibt:
Have tested this fix on the `Contour_extraction.ipynb` notebook and it works on both `rasterio` and Datacube-loaded datasets, so I think we can merge it in!

Issue #533, opened 2020-03-05
Update repository with recent DigitalEarthAfrica notebooks and code


robbibt:
In the recent ODC Hackathon 2020 there were some major additions and updates to code in the DEAfrica notebooks repository (https://github.com/digitalearthafrica/deafrica-sandbox-notebooks).

We should keep a list of relevant things that were updated in DEAfrica notebooks/code so that we can port them back over to `dea-notebooks`.

@cbur24, @caitlinadams, if you get a chance can you comment with any new notebooks/functionality that was developed during the hack that you think would also be useful to include in `dea-notebooks`? I'm thinking things like the `Dask` notebook I saw that would work just as well in the Australian context.

caitlinadams:
- We're most of the way through an introductory notebook to explain Dask: https://github.com/digitalearthafrica/deafrica-sandbox-notebooks/pull/38
- Potentially the Sentinel-1 notebook: https://github.com/digitalearthafrica/deafrica-sandbox-notebooks/blob/master/Datasets/Sentinel_1.ipynb
- Possibly the Urbanization case study I started: https://github.com/digitalearthafrica/deafrica-sandbox-notebooks/pull/37 -- However, I think this should probably be reworked to focus less on calculating a sustainable development goal, and potentially more on an Australian industry/monitoring kind of question

cbur24:
- The format/structure of the[ Landsat Collections](https://github.com/digitalearthafrica/deafrica-sandbox-notebooks/blob/refactor_loadard/Datasets/Landsat_collections.ipynb) notebook could work well for Aus.  Would need to switch out the USGS Landsat collections C1 & C2 for the Aus variants C3 & C2.
- [Wetlands Insight Tool](https://github.com/digitalearthafrica/deafrica-sandbox-notebooks/blob/master/Real_world_examples/Wetlands_insight_tool.ipynb) notebook under `Real_world_examples`.  Some adjustments would need to be made to the [deafrica_wetlandstools.py](https://github.com/digitalearthafrica/deafrica-sandbox-notebooks/blob/master/Scripts/deafrica_wetlandtools.py) script to make it work for Aus.

robbibt:
Another one I noticed come up in Github emails: was there a bug fix to `mostcommon_crs` to allow it to work when certain query/data load params are passed to it? @cbur24

cbur24:
Yes, that's right @robbibt . Pretty basic, I just removed the offending params:
```
[...]

PR #541, opened 2020-04-02
Porting Parallel processing using Dask notebook to dea-notebooks

robbibt:
### Proposed changes
This PR brings across the `Parallel processing using Dask` from DEAfrica Notebooks to DEA Notebooks (see #533). I've also updated any of our existing notebooks that tag Dask and added links in them that point to the new notebook.


robbibt:
@cbur24 @caitlinadams I just need to quickly test this on the NCI tomorrow after Gadi comes back online, but other than that it should be good to go

robbibt:
Confirmed that this works correctly on the NCI 👍

PR #542, opened 2020-04-03
Change basemap for filmstrips notebook

robbibt:
### Proposed changes
This PR changes the basemap used in the interactive map selector in the `Change_filmstrips.ipynb` notebook from the Google Maps/vector view to a satellite image basemap. This should make selecting areas to load satellite imagery for much easier.

Also made a minor update to the `dask_chunks` parameter to use a more sensible default.


robbibt:
Have run the notebook and it works well.

Also made a minor change to where the dask client is set up in the hope it might fix this issue: https://github.com/opendatacube/odc-tools/issues/34

PR #549, opened 2020-04-09
improve `xr_rasterize`

cbur24:
`xr_rasterize` failed when provided a 3D dataarray as a template due to the way it handled extracting coordinates, shapes, and dimensions.  This fix allows for handling 3D arrays.  As a result of this change, it can also handle passing either a dataarray or dataset as the template. I suggest testing this update in a workflow that attempts to mask a dataset using a geopandas geodataframe, eg:
```
mask = xr_rasterize(gdf, ds)
ds = ds.where(mask)
```
This fix should greatly simplify rasterizing polygons for masking (among other uses) as all of the crs reprojection, dimension/coordinate handling, and xarray construction is handled by `xr_rasterize`.


uchchwhash:
@cbur24 why not use `geobox.dims` for this?

cbur24:
Didn't know `ds.geobox.dims` was a thing! I will implement that in a try-except clause as not all datasets are going to have a geobox attribute

uchchwhash:
Thanks @cbur24! At the risk of sounding _really_ annoying now, it seems to me that you are going to have at least a `crs` object in hand. Why not use `crs.dimensions` then? :smile:

cbur24:
Cheeky @uchchwhash!  I updated the function to try and grab the `geobox.crs` before trying to grab the attribute crs (`da.crs`), and rearranged the order in which it grabs the dims and coords so the function can automatically know how to index for the coordinates.  But I'll leave out the `crs.dimensions` for now since there's a non-negligible chance that the crs will be a string rather than a `CRS` object, meaning it won't have the crs.dimensions attribute.

cbur24:
Great changes, @robbibt. Thanks for the review

PR #559, opened 2020-04-28
made a module for temporal tools for notebooks

BexDunn:
### Proposed changes
A module for time-based scripts is added to the scripts directory. These scripts are used in some of the WIT notebooks and a notebook for 'frequently used code snippets' that I'm hoping to push up today. 

I anticipate that we'll add more scripts to the module, hopefully around capabilities like 'find nearest timestep and retrieve data'.


robbibt:
I love the `assert` lines, we really should use more of them in code in this repo

PR #560, opened 2020-04-29
Add a Frequently used code notebook to demonstrate running an analysis on multiple polygons

BexDunn:
### Proposed changes
Add a frequently used code snippet that demonstrates loading and iterating through a shapefile with multiple polygons. This notebook is by request of a stakeholder.

- [x] Notebook created using the [DEA-notebooks template]
(https://github.com/GeoscienceAustralia/dea-notebooks/tree/develop)


cbur24:
My preference here would be to use `geopandas` rather than `fiona` for iterating through polygons, as this will make it consistent with the use of `geopandas` elsewhere in the repo (and we can then plot the polygons on a basemap using `map_shapefile`). Also, if we include masking in the workflow (as I would also suggest so its applicable for multiple applications), then the rasterizing step is simplified with `Spatial_tools/xr_rasterize`

BexDunn:
> My preference here would be to use `geopandas` rather than `fiona` for iterating through polygons, as this will make it consistent with the use of `geopandas` elsewhere in the repo (and we can then plot the polygons on a basemap using `map_shapefile`). Also, if we include masking in the workflow (as I would also suggest so its applicable for multiple applications), then the rasterizing step is simplified with `Spatial_tools/xr_rasterize`

I don't know how to do this using geopandas - I'm aware there are recent updates that make this easier, but I didn't think they're in datacube yet? I thought maybe if I got a notebook up, we can change it once the changes are in? Unless they're in and you know how to do this @cbur24 , I haven't done this with geopandas yet - I've just reproduced the workflow that underlies WIT on a simple scale for someone who asked how to do it. Be keen to learn how to do better faster etc.

..I was deliberately avoiding the masking step as I figured that this is just a 'Frequently used code' rather than a 'Real world example'. Thoughts? 

Please feel free to jump in and make changes @cbur24 if you'd like to fix!

cbur24:
The changes in `datacube 1.8` will make the `geopandas` option a little cleaner but otherwise it still works fine as-is, and would be trivial to update once 1.8 is released.  I will have a go at making changes and then we can see which we prefer / let others chip in with their thoughts.

BexDunn:
> The changes in `datacube 1.8` will make the `geopandas` option a little cleaner but otherwise it still works fine as-is, and would be trivial to update once 1.8 is released. I will have a go at making changes and then we can see which we prefer / let others chip in with their thoughts.

ok, makes sense! Thanks. I really should stop being so speciesist about pandas.

cbur24:
I haven't yet updated the markdown but the code and general layout for a geopandas based workflow is there to assess. Thoughts?

BexDunn:
> I haven't yet updated the markdown but the code and general layout for a geopandas based workflow is there to assess. Thoughts?

I like it, and given geopandas has so much more functionality it makes a lot of sense. Thanks for making the changes and explaining how to do it for me. I'll update the markdown and push up again for review. It looks a lot shinier now.

robbibt:
This looks great everyone! Let me know when I can check it out and make any changes without merge conflicting any of your work!

BexDunn:
> This looks great everyone! Let me know when I can check it out and make any changes without merge conflicting any of your work!

Should be good to go on my end!

cbur24:
I fleshed out a little more of the markdown section by adding descriptions for each of the `Analysis parameters`.  Ready for your review @robbibt

BexDunn:
> I fleshed out a little more of the markdown section by adding descriptions for each of the `Analysis parameters`. Ready for your review @robbibt

cool, thanks @cbur24 , I forgot those...

BexDunn:
> This looks fantastic, and fills a really important gap that we don't yet cover in the repo!
> 
> I made some extra additions to the markdown, and simplified some of the code (e.g. got rid of the `deepcopy` as it isn't strictly needed in this example, and used the `row` iterator object rather than indexing into the vector wherever possible.). Have also removed any last references to shapefiles only (replacing with "vector files" which includes GeoJSON).
> 
> @BexDunn : I think you might have been working from an older version of the DEA Notebooks template? It's changed a bit since then (particularly around the headers including logo and footers). :)

Thanks for pointing that out! I'll update my template. Still need to update my repo to the new version wholesale.. maybe tomorrow :)

robbibt:
@BexDunn I found it was easier to start a new/fresh branch as a copy of `develop`, and then slowly move across the files from my original branch as I needed them

PR #588, opened 2020-06-17
spatial wit - pending documentation

BexDunn:
### Proposed changes
This is a testing version of the Spatial WIT notebook. It only runs on the NCI at the moment. Once the documentation is complete, and collection 3 along with Fractional Cover and WOfS are available on the Sandbox, we'd like to bring it into the sandbox. 

-  ease of use
- documentation
- additions?

- [x ] Notebook created using the [DEA-notebooks template](https://github.com/GeoscienceAustralia/dea-notebooks/tree/develop)


emmaai:
Not sure if the output of cells are kept intentionally? But it's kinda difficult for future change.

LeoLym:
The current version of spatial WIT doesn't capture large WIT's along the same Landsat path.  Scenes along the same path are acquired seconds apart and can safely be combined, scenes from adjacent (east or west rather than north or south) rows can't be safely combined because they're acquired 8 days (or more) before/after that scene.

robbibt:
Hey @BexDunn , is this still a live PR? (just doing an early spring clean of our open PRs/issues!)

BexDunn:
> Hey @BexDunn , is this still a live PR? (just doing an early spring clean of our open PRs/issues!)

Hey @robbibt - it's not dead, just sleeping ;) This PR is waiting on progress on our documentation, and waiting on C3.

emmaai:
Can we merge this first so that we can keep track of change from PR #790 ? @BexDunn

BexDunn:
So, we need this PR approved in order to get the AWS spatial WIT in. This doesn't tick all the boxes, because we need the other PR on top of this plus changes.

PR #599, opened 2020-07-03
New notebook demonstrating pixel edge polygonisation

BexDunn:
### Proposed changes
New notebook demonstrating creating pixel boundaries as a vector files, for Landsat and Sentinel 2. Useful for Melbourne Water, ARI applications - happy to move to Scientific Workflows if you'd rather, but potentially also useful for drone imagery comparison - I believe it's multipurposeness means it should be good in Frequently used code. 


BexDunn:
> This looks great @BexDunn. I have a couple of small suggestions:
> 
> * can you keep the notebook outputs? It makes it easier to see what each cell is doing, particularly for example code like in `Frequently_used_code`
> * You can probably delete: 'Use standard import commands; some are shown below. Begin with any iPython magic commands, followed by standard Python packages, then any additional functionality you need from the Scripts directory.'. I think it's just a template comment, rather than a bit to publish?
> 
> Otherwise, it looks good!

Generally I would keep the outputs, except now it's a standard part of our PR that we dont...

CEKrause:
I just had a quick look through the other frequently used code notebooks, and they seem to have outputs in the repo?

CEKrause:
Wasn't removing the outputs just something you and Emma were doing? Maybe?

BexDunn:
Cool, will re-run the thing.

robbibt:
This looks awesome! Definitely a good one for `Frequently_used_code`. Let me know when you've pushed changes and I'll quickly review!

BexDunn:
> This looks awesome! Definitely a good one for `Frequently_used_code`. Let me know when you've pushed changes and I'll quickly review!

updated, potentially too late this evening :)

robbibt:
Looks great, I just made some minor formatting changes, as well as reducing the study area size so the notebook runs quicker.

My only slightly larger comment would be that it would be great to actually plot up some of the resulting shapefile results over the top of the `xarray` rasters directly inside the notebook, or at least include a screenshot of what the end result should look like so users don't necessarily need to download all the outputs and compare them in a GIS to see it. E.g. I think something like this both looks cool and really gets the point across about what the notebook produces:

image

The problem at the moment is that the area in the queries is too large to meaningfully plot the outputs... perhaps you could choose a super tiny area for each example (e.g. 20 x 20 pixels)?

BexDunn:
Those are excellent suggestions @robbibt - I'll cut down the size and have a crack at displaying them. Did you use the rgb with poly tools to display?

PR #607, opened 2020-07-17
Compositing notebooks and functions

cbur24:
### Proposed changes
Started redoing the `xr_geomedian` notebook and then decided to port over from DE Africa all things image compositing.  Two new notebooks: `Generating geomedian composites` and `Generating composites`; five new functions for `dea_datahandling`: `calc_geomedian`, `first`, `last`, `nearest`, `_select_along_axis`

I haven't run these on the NCI.


robbibt:
Reviewing now!

Scripts/dea_datahandling.py
```diff
@@ -704,3 +714,202 @@ def paths_to_datetimeindex(paths, string_slice=(0, 10)):
...
+    max_value : the maximum value that could be found in the array
+    min_value : the minimum value that could be found in the array
+    '''
+    da = (odc.algo.reshape_for_geomedian(ds, axis=axis) if
+          isinstance(ds, xr.Dataset) else ds)
+
+    if max_value is None:
+        max_value = da.max(skipna=True)
+    if min_value is None:
+        min_value = da.min(skipna=True)
```
Kirill888, line 744: So this will be really slow for dask inputs, it will essentially load all the data three times, one for max one for min and then again for actual computation.

Kirill888: also it is likely to be broken for `nodata=-999` case as it will pick `-999` as min values, as I do not think that `skipna` skips `nodata` values, only float NaNs

Kirill888: Scaling is only performed to improve numeric stability it is not a prerequisite for correct operation, so I would just skip scaling and convert to float with `*1+0` if user hasn't supplied min/max.

Also from API perspective maybe easier if that was something like `value_range=(min_val, max_val)` or some better name.

cbur24: For DE Africa, its handy having the `max_value` and `min_value` parameters as the satellite products have different surface reflectance scalings. Are the SR scaling values identical between all products for DEA? I think they might be (0-10,000), in which case we could remove the calculations and hard code the min and max values.

Scripts/dea_datahandling.py
```diff
@@ -704,3 +714,202 @@ def paths_to_datetimeindex(paths, string_slice=(0, 10)):
...
+    especially if using dask.
+
+    Parameters
+    ----------
+    da : xarray.DataArray object
+    max_value : the maximum value that could be found in the array
+    min_value : the minimum value that could be found in the array
+    '''
+    da = (odc.algo.reshape_for_geomedian(ds, axis=axis) if
+          isinstance(ds, xr.Dataset) else ds)
```
Kirill888, line 739: `xr_geomedian` does accept `xr.Dataset` objects now and performs reshape internally.

Scripts/dea_datahandling.py
Kirill888: also checkout `int_geomedian`, I think that gives better performance as it delays conversion to float until very end, just before calling to Dale's code, so it has lower memory and bandwidth requirement.

Scripts/dea_datahandling.py
cbur24: Thanks @Kirill888, can you update the code in this PR to reflect this change? And then I will port it across to DE Africa.

Scripts/dea_datahandling.py
Kirill888: Have you observed significant difference of output quality when selecting "exact" scaling factors vs none? 3 times slower operation and 3 times bandwidth better be worth it.

Kirill888: Besides we are not preventing that use case, just changing default behaviour to be less "surprising". We can just document the fact that scaling can help somewhat and if you know rough range of the data you can supply it and maybe get slightly better results out of it. And if not -- nothing bad.

With code as is, it is broken when given completely uniform input, as is will attempt `/0`

Scripts/dea_datahandling.py
Kirill888: I think `odc.algo.int_geomedian` is exactly this function, but has different internal implementation that takes advantage of knowing that input is `int`.

`int_geomedian` does assume `time,y,x` dimension order, but all of our data is like that, and instead of min/max it takes scale offset with default being no scale no offset

robbibt:
Hey @Kirill888 and @cbur24 , I currently have the branch checked out and am making changes, so I'm happy to refactor the code RE Kirill's suggestions above if that works. Just so I have it right: we can probably replace the entire `calc_geomedian` func with `odc.algo.int_geomedian` instead?

Scripts/dea_datahandling.py
Kirill888: hold on, in the example notebook loaded data is already `float32` cause this happened inside `load_ard`

robbibt: Ah yes, if the data was loaded in `load_ard` with cloud masking selected it will be in `float32`

robbibt: I think we should be able to get it back as `Int` if we add `dtype='native'` to the function call

Kirill888: I reckon staying in `int16` throughout is preferable.
`xr_geomedian` for float inputs, `int_geomedian` for int16 inputs, ideally I should merge it into one function though.

cbur24:
@robbibt if you could incorporate @Kirill888 suggestions that would great.  It seems like we probably can remove the `advanced control for geomedian` part.  Maybe we could instead show the use of `int_geomedian` and `xr_geomedian`? With a few words about how they differ? Something about speed and memory vs the desired output?

Kirill888:
yep what @cbur24 said above.

Needs a note about scaling factors: basically aim for roughly 0-1 range after scaling, but it doesn't need to be precise and can always default to operating on original values.

robbibt:
Hey @cbur24 and @Kirill888 , I've made a first pass at re-writing the notebook. I've changed up the order a fair bit for a simpler/more logical flow (load data > int_geomedian > grouping and resampling). `int_geomedian` works nicely after changing `load_ard` to `dtype='native'`, so I've added a bit up the top about memory use and why keeping the data in Int is good.

That said, I don't think I have enough knowledge about the workflow required to use `xr_geomedian` to write about it in the notebook. I don't know what parts of the "advanced" section are required as a bare minimum example, so I've had to leave it out for now. I think if either of you wanted to cover it, putting it in as a final section would be good so users aren't scared away by scaling factors etc earlier on.

Scripts/dea_datahandling.py
robbibt: Having them as one function would be really nice, `xr_geomedian` is a nice name for a general purpose wrapper function I think

robbibt:
I've also had a quick run through of the non-geomedian composite notebook and made some minor changes. I really like it, but there's something very strange going on in one of the plots:

image

That looks to me like possibly something confused with axes being plotted in the wrong order... @andrewdhicks?

Kirill888:
@robbibt thanks, this looks good. `xr_geomedian` is basically the same as `int_geomedian` except it assumes that input data is already in `float32` format. It can also operate directly on `xr.DataArray` and even on numpy arrays. But most should interface with it using `xr.Dataset`. 

When input is a single array the dimensions ought to be in this order `y, x, band, time`, with `time` dimension being removed by the process. This is awkward order, but that's what Dale's code expects natively, I guess it optimizes for computation performance.

I think the whole scale/offset can be ignored in the notebook, I do need to document it better in the function doc comment though.

cbur24:
@Kirill888 I've added a section at the end of the geomedian notebook demonstrating the use of `xr_geomedian`, if you could please give it a quick look and see that it makes sensethat'd be great.  Overall, I think the notebook is looking really good now. I've also removed the `calc_geomedian` code from `dea_datahandling`.

@robbibt I think I resolved the 'age of pixels' plot. The issue was with this line:
```
# Identify last time in our data
last_time = last_blue.time[-1]
```
Identifying the 'last' time isn't what we wanted, we wanted the _latest_ time, because then we can compare each pixel in the 'last_blue' composite with the lastest time.  ie
```
# Identify lastest time in our data
last_time = last_blue.time.max()
```

Kirill888:
actually I default `num_threads=1` in my code, so doesn't matter what Dale's defaults are.

cbur24:
Thanks @Kirill888.  I placed `num_threads=1` in there because without it the `xr_geomedian` code printed "working on 16 threads", wasn't sure if that was just an errant print statement or if it was having an impact, either way it was confusing so I think its fine to leave in there for now.

Kirill888:
looks like it's only `int_geomedian` that sets the default value to `1`, I should change `xr` version to do the same.

robbibt:
@cbur24 Just a small typo in the composite notebook: 
image

Any way we can remove the FutureWarning by adding the param they recommend?

PR #619, opened 2020-08-17
Add write cog and data load exceptions to Burnt area mapping notebook

Gabzgit:
### Proposed changes
- Write_geotiff has been deprecated, use write_cog instead
- Make edits to accommodate when no NRT is available
- Something appears to have made less Sentinel 2 data available around the originally chosen dates, change dates to include more post-fire images

NOTE: write_cog was not working for me even though I removed existing files. Not sure how to fix.

- [ /] Notebook created using the [DEA-notebooks template](https://github.com/GeoscienceAustralia/dea-notebooks/tree/develop)
- [/ ] Remove any unused Python packages from `Load packages`
- [ /] Remove any unused/empty code cells
- [/ ] Remove any guidance cells (e.g. `General advice`)
- [/ ] Ensure that all code cells follow the [PEP8 standard](https://www.python.org/dev/peps/pep-0008/) for code. The `jupyterlab_code_formatter` tool can be used to format code cells to a consistent style: select each code cell, then click `Edit` and then one of the `Apply X Formatter` options (`YAPF` or `Black` are recommended).
- [ /] Include relevant tags in the final notebook cell (refer to the [DEA Tags Index](https://docs.dea.ga.gov.au/genindex.html), and re-use tags if possible)
- [/ ] Clear all outputs, run notebook from start to finish, and save the notebook in the state where all cells have been sequentially evaluated
- [ /] If applicable, update the `Notebook currently compatible with the NCI|DEA Sandbox environment only` line below the notebook title to reflect the environments the notebook is compatible with


BexDunn:
Hi @Gabzgit  - I had the same problem with the data - not sure what's happening. Might be worth chasing down? I have some ideas about write_cog, but I didn't have any luck checking out the branch on the sandbox and pulling - something weird is happening? Looks like I've got an old version of the notebook without the write_cog changes even when pulling the `add_cog` branch.

BexDunn (approved):
OK, I've re-checked this, re-ran, and fixed the write-cog issues. I think it's all good.

PR #622, opened 2020-08-31
Prevent `rgb` from doing multiple dask computes over the same data

robbibt:
### Proposed changes
@andrewdhicks found a bug in DEAfrica where the `rgb` function makes multiple passes over the same data when using `dask` when `robust=True` to internally calculate the 2 and 98 percentiles:

https://github.com/digitalearthafrica/deafrica-sandbox-notebooks/issues/141#issue-664126942

This PR fixes this issue by specifically computing the array of data to be plotted earlier in the process, so the percentile calculation does not need to compute it a second time.

### DIFF
--- a/Scripts/dea_plotting.py
+++ b/Scripts/dea_plotting.py
@@ -23,7 +23,7 @@ Functions included:
     map_shapefile
     xr_animation
 
-Last modified: June 2020
+Last modified: September 2020
 
 '''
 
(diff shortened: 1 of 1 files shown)


cbur24 (approved):
Looks good to me, @robbibt. I ran the code through the composites notebook and it computed the arrays only once and plotted perfectly.

Issue #586, opened 2020-06-09
update exporting geotiff notebook to use write_cog


BexDunn:
write_geotiff has been deprecated by write_cog in datacube 1.8.0 https://datacube-core.readthedocs.io/en/latest/about/whats_new.html 

- update the exporting geotiff notebook
- maybe also update the script if it needs it?

caitlinadams:
Chad's initial notes on what needs to be done:

1. Use the info here: https://datacube-core.readthedocs.io/en/latest/dev/api/generate/datacube.utils.cog.write_cog.html to re-write the Exporting_GeoTIFFS.ipynb file: `from datacube.helpers import write_geotiff` function is being replaced by the datacube.utils.cog.write_cog function.  Be sure to make a note of the extra memory limitations of this function, and leave the code block on using the array_to_geotiff function so a user can decide not to use the write_cog function if they prefer not to use the extra memory.
2. Go through the Frequently_used_code and Real_world_examples folders and replace any write_geotiff code blocks with write_cog . 

Any other recommendations for what specifically needs to be done appreciated!

BexDunn:
Additonally please explicitly pass dtype and nodata from the rasterio options in

robbibt:
I might have a go at this this week if that's OK?

PR #623, opened 2020-09-01
Update Exporting GeoTIFF notebook to use `write_cog`

robbibt:
### Proposed changes
Code in this repo needs to be updated to use the newer `write_cog` function instead of the deprecated `write_geotiff` (e.g. #586). This PR starts this process by updating the "Write_GeoTIFF" notebook to use `write_cog`, so at least we have some nice examples to point new users to.

Have also included two brief sections on:
1. Supplying custom `rasterio.open` kwargs like nodata
2. Using `write_cog` with `dask` arrays

Still to do (in a second PR): replace all instances of `write_geotiff` in other notebooks with `write_cog`.

### Closes issues (optional)
Partially contributes to #586


robbibt:
@Kirill888 No need for a detailed review, but just bringing to your attention in case anything about `write_cog` is clearly incorrect

Kirill888:
This looks good, thanks. Only other thing maybe is to talk about exporting `float32` data with `nan` that should not have `nodata` set at all. To achieve this one has to make sure that `nodata` attribute is not set on the array, there is no way to overwrite `nodata` into "nothing", so one has to `xx.attrs.pop('nodata', 0)` if they want to generate COG without `nodata` set. 

Other comment would be around `.isel(time=1)`, so this saves the second slice, but can be confused for saving "first" slice instead. So maybe `.isel(time=0)` is more copy-paste friendly, and if first slice is too boring maybe pick something other than `1`, so that it's more obvious that this needs editing after pasting.

robbibt:
Thanks @Kirill888, those are great suggestions. I'm just starting to write up a bit on your first point re nodata and float arrays... did you have a particular scenario in mind for this? Just thinking that it's still technically possible to have a float array with `NaN` values but also a nodata attribute, for example, you mask out clouds with `NaN`, but there are still -999 values in the data that haven't been replaced (so the attribute is still valid)?. I also was writing up the example based on when you've used `mask_invalid_data` to set nodata values to `NaN`, but it seems that function already drops the `nodata` attribute automatically so the user doesn't need to do it themselves.

Kirill888:
@robbibt I think someone Bex collaborates with had this problem, one of index computations utilities didn't drop `nodata` flag from output and they were confused as to why `nodata=0` was set on the output file. It's really a limitation, a gotcha, of the API, one can replace `nodata` with some other value at call time but one can not request "to not write one at all", since if you just supply `nodata=None` this really means "use nodata flag on the input and if set write that to file" rather than "do not set nodata flag on the output image at all regardless of what input is".

Regarding floats, yes you can have both NaN and some other special value also, but I would say this should be discouraged, whole idea of `nodata` is to emulate NaN for data stored in integer formats. If you use float data, you have NaN, so use that.

BexDunn (approved):
Hey Robbi. The bit on nodata looks both good and useful to me, and I think it's handy to have the commentary on dask as well. Nice work.

PR #626, opened 2020-09-11
Better document `inplace` functionality for `calculate_indices`

robbibt:
### Proposed changes
Currently the `calculate_indices` function from `dea_bandindices.py` deep copies the input data to a new object in memory. While this prevents unexpected behaviour (e.g. users being confused if their input dataset was modified inplace when they expect only their newly assigned variable to have new bands), it is twice as costly in terms of memory use which makes it not an ideal choice for large memory analyses.

The function already has a `deep_copy` param which can be set to `False` to allow the user to assign new variables to their original data without creating a new object in memory. However, this feature was poorly documented in the code and in the notebook.

This PR:
- Renames `deep_copy` to `inplace` for consistency with `pandas`
- Improves documentation of this param in docstring
- Added a new section discussing this param in the notebook
- Fixes a bug where bands were not dropped when `deep_copy=False`


MatthewJA (approved):
LGTM

Issue #618, opened 2020-08-13
Update DEA Notebooks template and notebook headers to contain a direct Sandbox link 


robbibt:
In meeting with Today regarding the new DEA website, it was suggested that we should include an easy link to every example notebook that will take users directly to the DEA Sandbox so they can run code interactively.

This would be easy to do by updating the DEA Notebooks header template, ideally via a bulk update:

https://www.internalpointers.com/post/linux-find-and-replace-text-multiple-files

MatthewJA:
I think we can close this now, right @robbibt?

robbibt:
Yep, closing

PR #636, opened 2020-10-07
Add links to Sandbox to notebook headers

robbibt:
### Proposed changes
During Today's redesign of the DEA website, it was discussed that including a link to the Sandbox within each notebook in this repo would be valuable for allowing users to go directly from DEA documentation to interactively explore the code: #618 

This PR adds a new "Sign up to the DEA Sandbox" link to the top of each notebook to make it easier for users to access DEA Sandbox:
image

(I've also made some small updates to the two new Beginners guide notebooks to use PEP8)

### Closes issues (optional)
- Closes Issue #618


BexDunn (approved):
Looks awesome and works. I notice that it doesn't seem to be a change on the Scientific workflows notebooks - is this deliberate? What did you use for the find in the find and replace?

BexDunn:
Oh, I noticed that you may have missed adding this to the template? Shall I make this a new issue @robbibt or do you want to update this one?

MatthewJA:
LGTM if you add it to the template too :)

robbibt:
Thanks @BexDunn and @MatthewJA , have added to the template! 

RE Scientific workflows: those notebooks aren't currently synced to the Sandbox, so I left the message out from them for now as users might get confused if they can't find the notebook they clicked from

MatthewJA:
@robbibt If that's the case then it might be a good idea to add a comment to the template that the sandbox link should only be added for sandbox-synced notebooks?

robbibt:
@MatthewJA Done! And removed some old instructions about choosing an area that are no longer relevant

MatthewJA (commented):
Looks great!

"Compatibility" is spelled wrong in the new line. Though it seems to be consistent with the rest of the notebooks, which also mostly spell compatibility wrong. Is that something for a new PR?

robbibt:
Haha yep, let's save that one for another issue

MatthewJA (approved):
LGTM :)

Issue #640, opened 2020-10-09
Add more virtual product examples (or even a bunch of virtual products)


MatthewJA:
I found it really hard to make a virtual product that did what I wanted it to do. The examples listed in the relevant notebook _work_, and are _great_, but deviating from them a little can break with very cryptic errors. I'd love either more examples, or possibly just a collection of ready-to-go virtual products (e.g. combined Sentinel 2, masked LS8...).

robbibt:
Yes, this would be really cool. For now I think the examples should live here so we have them all in one nice central catalogue:

https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Supplementary_data/Virtual_products/virtual-product-catalog.yaml

These ones would be particularly worth focusing on I think, as they could potentially be used for real (vs. demo) analyses:

- Combined Landsat 5, 7 and 8
- Combined Landsat 5, 7 and 8, cloud masked
- Combined Sentinel-2A and 2B
- Combined Sentinel-2A and 2B, cloud masked
- Combined Landsat 5, 7, 8, Sentinel-2A and Sentinel-2B
- Combined Landsat 5, 7, 8, Sentinel-2A and Sentinel-2B, cloud masked

robbibt:
I'd also love to have some extra documentation/comments in that YAML that separate the examples into the "demo" style recipes and the analysis-style recipes. I guess alternatively we could have one YAML called `virtual-product-examples.yaml` for the ones that already exist, and `virtual-product-catalog.yaml` for the more complex analysis-style ones?

MatthewJA:
I'd kinda like multiple YAML files - a monolithic YAML file is hard to navigate and understand, as I discovered when I tried to learn about virtual products the other day.

robbibt:
Yep OK, that's fair enough. I guess the question is how granular we want to go e.g. broken down to individual products/product suites? (e.g. surface reflectance virtual products, WOFS virtual products etc)? I don't mind having a few if they're nicely named and live in an easy to navigate to location like that "Virtual_products" folder

robbibt:
I have a Landsat 5, 7 and 8 combined virtual product available here that could be handy to add: https://gist.github.com/robbibt/5e1bf4e63d1f429ca2ee078227801dfc

Also pinging @uchchwhash who probably has some thoughts on this stuff too

uchchwhash:
What can I do to help? :smile: 

[...]

PR #641, opened 2020-10-14
Sentinel 2 waterbodies notebook

MatthewJA:
Here's a notebook that generates polygons with Sentinel 2 MNDWI, Landsat 8 MNDWI, and WOfS, and compares their time series.

I haven't a) run it in sequential order, or b) run it on the NCI, because it takes ages to run and I want to make sure it's feature complete first.


robbibt (changes requested):
Hey @MatthewJA , another super cool notebook. I have some overall comments below - I think the main things that need to be fixed up are improving code and markdown documentation and perhaps thinking about ways to simplify the notebook (e.g. removing any sections which aren't absolutely critical to the overall analysis). 

- If the notebook takes a while to load (more than a few minutes), it would be good to focus on a smaller example area (e.g. a single waterbody). That way it can serve as a nice quick to run demo, but still allow users who want to analyse a larger area to modify the extent and re-run.
- Some sections need additional markdown explanation and code comments to walk the user through what each piece of the analysis is doing. In particular, "Build a virtual product for the MNDWI calculation" (as many users will not have loaded data using virtual products before) and "Surface area time series calibration", "Generating time series", "Comparing time series" (anywhere there's more than a line or two of code, or functions without docstrings... ideally we want reasonably inexperienced users to be able to follow every step of the analysis, or at least appreciate why it's being done).
- The "cliquing" section should have some additional explanation aimed at explaining what this does and why it is useful for users who have never used this. Same for the graph at cell 28; some extra explanation on how to interpret the graph aimed at new users would be great.
- The "What do the time series look like?" is currently very long. I would choose a single example graph rather than a large sequence or a handful of contrasting results that can fit inside an single figure, and then explain just those in more detail.
- We are currently avoiding custom pip installs in these example notebooks as they reduce reproducibility and lead to downstream issues where users come to us complaining about issues based on a customised environment, which makes it difficult to bug fix. Is it possible to replace the functionality in `pip install rioxarray disjoint-set` with functionality from existing packages? Otherwise we can try and get the new package installed, although for `rioxarray` it might be a bit tricky as that package has a large amount of overlap with `datacube` itself...
- Would a conclusions section here be useful? There's a lot of different things going on, and it might be nice to wrap things up neatly at the end.

Minor updates:
- Weird formatting on the square metre units here - does this use markdown formatting or something else?
image
- Move "analysis parameters" below "Connect to the datacube", and ensure all sections follow order in template notebook
- Split `:
bbox = xlim, ylim = ((143.70774, 143.89147), (-35.71674, -35.48853))  # Kerang Wetlands` into two lines to match other notebooks and drop `bbox =` if it's not being used
- Update `dc = datacube.Datacube(app="Sentinel2Waterbodies")` to `dc = datacube.Datacube(app="Sentinel_2_Waterbodies")` to match notebook title
- Change to direct link: "MNDWI (DOI:10.1080/01431160600589179)"
- Can `mask_to_polygons` be replaced with the existing `xr_polygonise` function? 
- Can we replace the `%%bash` block with loading the Waterbodies data directly using `pandas`? This would reduce the amount of new techniques users need to learn, and bring it in line with the existing approach in `Surface_water_duration.ipynb`.

robbibt:
The virtual product is super cool by the way. I think it probably works fine in-text here for now, but in time we should think about moving it out to a separate YAML, as per #640

MatthewJA:
Thanks for the feedback @CEKrause and @robbibt. Super helpful for getting this notebook to where I want to get it. I'll work on it and get back to y'all.

MatthewJA:
Thanks again, super helpful. I've stripped it down and I think it makes a bit more sense now, and I've tidied up the bits requested.

I did have an issue with Dask where it was _ridiculously_ slow without preloading, so could someone please take a look at my Dask code to see what I'm doing wrong? (If anything, I guess. Preloading seems to fix it just fine)

robbibt (approved):
Have pushed some updates below, mainly to add some extra code documentation and to reduce the time period further to reduce the load time. The dask issue is annoying, but we can resolve this outside this PR

PR #648, opened 2020-10-20
fixed broken header but forgot to rerun for version and update info, …

BexDunn:
rerun the updated notebook with version and update info after I accidentally pushed the fix to develop...


robbibt (approved):
Looks great! 👍

Issue #545, opened 2020-04-06
Add polygon drill notebook into new master


CEKrause:
master-archive - 05_Temporal_analysis/PolygonDrill.ipynb

PR #649, opened 2020-10-20
Add polygon drill to repo

CEKrause:
### Proposed changes
The polygon drill notebook was never ported across from the master_archive branch. 

### Closes issues (optional)
- Closes Issue #545 


robbibt (approved):
Looks great, can't believe we've gone so long without a nice example of this in the repo!

I've checked out the branch and made some small changes: the main one was to change the plots from plotting a single time step to plotting the full time series, which hopefully gives a better sense that the polygon drill is returning a time series of imagery through time. 

Also it blows my mind that you can use geopandas to load zipped data directly from a URL!

CEKrause:
I was pretty chuffed to discover that you can pass geopandas a URL. It makes life so much tidier!

MatthewJA (approved):
LGTM too! Super useful notebook to have around.

PR #651, opened 2020-10-22
Added climate driver notebook

MatthewJA:
Added a notebook that estimates climate driver influence on rainfall. I want to use this as a comparison for doing the same thing with waterbodies, since the behaviour of rainfall is better understood than the behaviour of wet surface area.

### DIFF
--- a/Real_world_examples/Estimate_climate_driver_influence_on_rainfall.ipynb
+++ b/Real_world_examples/Estimate_climate_driver_influence_on_rainfall.ipynb
@@ -57,7 +57,8 @@
     "import mpl_toolkits.basemap\n",
     "\n",
     "import sys\n",
-    "sys.path.insert(1, '../Scripts')\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


MatthewJA:
Only works on the NCI, since the BOM rainfall grid isn't available on the sandbox.

CEKrause (changes requested):
Few small comments:
- the DEA logo doesn't seem to be loading properly in the document header. Maybe the file path needs to be modified?
- Can you please add in some more code comments to indicate what you are doing in each cell? I pretty much comment every line of code, especially for things being pushed to develop where people unfamiliar with what you're doing will see it. 
- Super cool code workflow being shared here.

MatthewJA:
@CEKrause the logo seems to be working fine when I load this up on Sandbox...
image

CEKrause (approved):
Woo. Nice additional comments and code doco, thanks.

PR #652, opened 2020-10-22
Fix major scaling bug in xr_animation, remove deprecated funcs

robbibt:
### Proposed changes
Lan-Wei encountered a nasty issue where one-band animations were being plotted with unexpected colours that did not match the underlying data when using the default `percentile_clip`. This revealed that the animation code was inadvertently plotting colour stretches for each frame independently, which meant strictly that animation frames could not be compared fairly across time. This was not noticed up to now as most animations are based on multi-band satellite data with approximately similar distributions in each frame - Lan-Wei's example involved plotting single-band timesteps with a single unique value in each frame, which made the scaling issue much more obvious.

A secondary issue involved a scenario where the user manually specifies a `vmin` and `vmax` used to scale colours. As part of the underlying code, those values are also used to scale all array values to between 0 and 1.0 before plotting the data. These 0-1.0 rescaled values are then passed to `plt.imshow` - however, the code accidently also passed the original `vmin` and `vmax` values and used those to set the colour stretch. This is incorrect: once the data has been rescaled to between 0 and 1.0, the original `vmin` and `vmax` values no longer apply. Instead, each frame should always be plotted with a colour stretch between 0 and 1.0.

I've fixed both of these bugs by "popping" the original `vmin` and `vmax` out of the dictionary of kwargs that are used to customise `plt.imshow`, and forcing `plt.imshow` to always use a `vmin` of 0.0 and a `vmax` of 1.0. This ensures that the colours match the rescaled data, and are consistent across all animation frames. 

## Additional changes
Since the outdated `animated_timeseries` has now been depreciated for almost a year, I've also finally removed these funcs from the script. If users complain about this removal, we can point them to the `xr_animation` function and accompanying notebook.

### DIFF
--- a/Scripts/dea_plotting.py
+++ b/Scripts/dea_plotting.py
@@ -765,8 +765,7 @@ def xr_animation(ds,
         animate each frame in the animation. Plots array and any text
         annotations, as well as a temporal subset of `gdf` data based
         on the times specified in 'start_time' and 'end_time' columns.
-        """
-        
+        """        
 
         # Clear previous frame to optimise render speed and plot imagery
         ax.clear()
@@ -862,7 +861,7 @@ def xr_animation(ds,
 
     # Convert data to 4D numpy array of shape [time, y, x, bands]
     ds = ds[bands].to_array().transpose(..., 'variable')[0:limit, ...]
-    array = ds.values
+    array = ds.astype(np.float32).values
 
     # Optionally apply image processing along axis 0 (e.g. to each timestep)
     bar_format='{l_bar}{bar}| {n_fmt}/{total_fmt} ({remaining_s:.1f} ' \
--- a/Scripts/dea_plotting.py
+++ b/Scripts/dea_plotting.py
@@ -23,7 +23,7 @@ Functions included:
     map_shapefile
     xr_animation
 
-Last modified: September 2020
+Last modified: October 2020
 
 '''
 
@@ -568,7 +568,7 @@ def xr_animation(ds,
     Supports .mp4 (ideal for Twitter/social media) and .gif (ideal 
     for all purposes, but can have large file sizes) format files. 
     
-    Last modified: April 2020
+    Last modified: October 2020
     
     Parameters
     ----------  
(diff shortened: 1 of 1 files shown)


robbibt:
@cbur24 Tagged you as a reviewer as this code probably needs to be updated for DEAfrica too, otherwise strictly all animations produced using `xr_animation` potentially give misleading results

cbur24:
> 
> 
> @cbur24 Tagged you as a reviewer as this code probably needs to be updated for DEAfrica too, otherwise strictly all animations produced using `xr_animation` potentially give misleading results

Thanks @robbibt will take a look this afternoon

cbur24 (approved):
@robbibt I ran these code changes through both the DEA and DEAfrica versions of `animated_timeseries.ipynb` notebooks without any hiccups so happy to approve.  One thing I noticed that I think is unrelated to these changes: the final plot in the DEA animation notebook (custom image processing) returns entirely blank frames. I thought this was due to no-data values messing with the scaling functions but passing `ds.fillna(0)` didn't resolve the issue. This might be worth investigating more and/or removing that code from the notebook.

robbibt:
Hmm, good catch! This makes me think something else weird is going on - will investigate.

robbibt:
Thanks @cbur24 , that was another bug: the function converts the xarray data to a numpy array, then applies the `image_proc_funcs` to each timestep in this numpy array before overwriting the original data. The problem occurs when the input data has an integer dtype, but the `image_proc_funcs` output data with a small floating dtype:
* User inputs data with integer values from 0 to 10,000
* A `image_proc_funcs` function processes this data and outputs values between 0.0 and 1.0
* These are written back into the original array with a integer dtype. Because almost all values are now below 1 (the smallest int), the output array comes out black in the animation.

Have fixed this by forcing all arrays into float32  when they are converted to a numpy array. This shouldn't have any negative impacts on memory use, as the data is converted to floating points anyway prior to plotting in a subsequent step.

cbur24:
@robbibt nice piece of investigation! I will port this all over to De Africa as well

PR #655, opened 2020-10-23
Added verbose flag to subpixel_contours

MatthewJA:
Tiny change: added a verbose flag to `dea_spatialtools.subpixel_contours` so you can suppress the debugging messages.

Sorry for the email spam. >.>


robbibt (approved):
I like it!

PR #657, opened 2020-10-26
Made exceptions more specific in subpixel_contours

MatthewJA:
Non-specific exceptions are hard (read: not really possible) to catch, and therefore discouraged in Python. I've changed all the `Exception`s in `subpixel_contours` to `ValueError`, `TypeError`, or `RuntimeError` as most appropriate. I also made it so that no-contours doesn't raise an error unless `errors="raise"`.


robbibt (approved):
Really nice, thanks for adding this in!

PR #662, opened 2020-10-27
Added verbose flags and different exceptions to spatialtools

MatthewJA:
This adds some verbose flags and more specific exceptions to the other major functions in `dea_spatialtools`, which I did for `subpixel_contours` in #657 and #654 but didn't realise needed to be done for the other functions too. The final three functions seemed OK without a flag. `xr_vectorize` doesn't use the verbose flag as it stands but I've included one for API consistency.


robbibt (approved):
Nice!

PR #675, opened 2020-10-28
adding urban change

cbur24:
### Proposed changes
A new `Real_world_examples` notebook that examines urban area through time to generate a map or urban-growth-hotspots


Gabzgit (approved):
Hello Chad,
This looks great. I only change lat lons from Africa to Australia so the markdown matched the area loaded.

I only have one question which I don't think really matters. Is the histogram the combination of the two geomedians, and so it's just showing all the pixels and providing a visual for where the threshold will go? Wasn't sure if a histogram per year would be worth it, otherwise it seems like most pixels would just be doubled up (except for small percentage of extra urban in the following year).

robbibt:
This is such a cool notebook - can't believe how nicely the new areas of development pop out

@Gabzgit : looks like I was making some edits at the same time as you, I think I fixed the merge conflict but let me know if it looks like something went wrong!

PR #676, opened 2020-10-28
Add ERA5 notebook to Australia repo

Gabzgit:
### Proposed changes
Add the ERA5 notebook from Africa to Australia repo
Add the script for importing ERA5

- [ x] Remove any unused/empty code cells
- [ x] Remove any guidance cells (e.g. `General advice`)
- [ x] Ensure that all code cells follow the [PEP8 standard](https://www.python.org/dev/peps/pep-0008/) for code. The `jupyterlab_code_formatter` tool can be used to format code cells to a consistent style: select each code cell, then click `Edit` and then one of the `Apply X Formatter` options (`YAPF` or `Black` are recommended).
- [ ?] Include relevant tags in the final notebook cell (refer to the [DEA Tags Index](https://docs.dea.ga.gov.au/genindex.html), and re-use tags if possible)
- [ x] Clear all outputs, run notebook from start to finish, and save the notebook in the state where all cells have been sequentially evaluated
- [ x] Test notebook on both the `NCI` and `DEA Sandbox` (flag if not working as part of PR and ask for help to solve if needed)
- [ x] If applicable, update the `Notebook currently compatible with the NCI|DEA Sandbox environment only` line below the notebook title to reflect the environments the notebook is compatible with

*The tags are the same for Africa nb, not sure if we need to change the tags or update the pages that says what tags are allowed


cbur24 (approved):
Great notebook @Gabzgit. Again, I just reduced the time range to a single month to speed up the run-time and reduce the amount of data downloaded.  I also renamed the csv notebook to have the same naming convention as this one (i.e. `External_data_<data-type>.ipynb`)

robbibt:
Another really cool notebook - this is perfect as it is, but in the future it would be great to combine this workflow with data from the datacube to demonstrate how climate data can be combined with satellite data to add value. 

I've added the renamed notebooks to the readme file so they appear in the user guide, and renamed the ERA5 script to be more generic "dea_climate.py" so we can use it for other climate datasets in the future.

Added a `no_testing` tag so it doesn't get run by future automatic testing due to the amount of data being downloaded.

PR #678, opened 2020-10-28
Add updated DEA High and Low Tide Imagery notebooks

### LINKED ISSUE #669: Update DEA_notebooks folder to include High/Low Tide Composite
From Issue #615.

robbibt:
### Proposed changes
This PR adds a simple DEA Datasets notebook for the DEA High and Low Tides Imagery product suite.

I've tried to keep this really simple and direct, base the text directly from the [CMI listing](https://cmi.ga.gov.au/data-products/dea/133/dea-high-and-low-tide-imagery-landsat), and to cover the following steps:

- Inspecting the products and measurements available in the datacube
- Loading low and high tide data for a coastal location
- Plotting low and high tide data in true and false colour
- Converting low and high tide data to a remote sensing water index (NDWI), and use this to map wet pixels at low and high tide

### Closes issues (optional)
- Closes Issue #669 


robbibt:
Can whoever reviews this please also critique this as a potential structure/template that we could use for other DEA Datasets notebooks? It would be fantastic if all of the notebooks in this folder could follow a reasonably similar structure so it's easy for users to get started with all of our different products.

MatthewJA:
> Can whoever reviews this please also critique this as a potential structure/template that we could use for other DEA Datasets notebooks? It would be fantastic if all of the notebooks in this folder could follow a reasonably similar structure so it's easy for users to get started with all of our different products.

I wonder if each notebook should stand individually (explaining datacube, list products, etc) or whether reading the beginner's guide should be assumed.

MatthewJA (approved):
LGTM, nice notebook. Simple and to the point with a nice example at the end. Runs real fast.

robbibt:
@MatthewJA I think we can probably assume users have gone through the Beginner's Guide (we could always add this as a pre-req at the top of DEA dataset notebooks to make this nice and clear?). My intention for including the list products bit was to basically explain what products are available via code, rather than listing them directly in the markdown (e.g. and give users strategies they can use themselves to find relevant products). Not sure if it will be too repetitive though?

MatthewJA:
I think it's good to include list products here, I just found the description wordier than I perhaps expected. But it's totally fine as it stands

PR #688, opened 2020-10-30
Add notebook. Notebook updated from MDBA tag in this repo

CEKrause:
### Proposed changes
This notebook was originally written by @katedugdale. I have updated it for inclusion in the stable branch of dea-notebooks. 


robbibt:
Hey @CEKrause... could this be a similar notebook to the existing `Inundation_mapping.ipynb` from `Real_world_examples`? 

https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Real_world_examples/Inundation_mapping.ipynb

MatthewJA:
It does look pretty similar. If it's different, maybe those differences could be exaggerated?

CEKrause:
Aw man! I didn't realise this other one was there. My new one is simpler and has more code documentation. Maybe it can replace `Inundation_mapping.ipynb`?

MatthewJA:
Yeah, the new one looks way better - the interactive gauge thing is cool. The old one also uses a pickle, which I'm pretty hard against as a way to store data... I'm quite happy with this one replacing the old one.

BexDunn (approved):
Looks good to me. I'm okay with keeping both, but if you want to remove the Frequently used code one, that's fine too :)

MatthewJA:
Got a few minor comments I've added, mostly suggesting the removal of some of the placeholder text and making a few things clearer

robbibt:
From memory, the pickling step was added by Kirill and Kate to serve as a cache as the BOM service goes down so regularly that the notebook was unreliable without some kind of backup option. This bit below was an important addition as it made sure the map only displayed only stream gauges that contained valid data (otherwise it can be challenging to find useful locations on the map). We could dump the pickled station data and see if can get away with just using `dea_bom.get_stations()`, but I'd like to keep the valid data filtering below if possible.

```
# Filter list to stations with available data
stations_with_data = pickle.load(open(str('../Supplementary_data/Inundation_mapping/stations_with_data.pkl'), 'rb'))
stations = [i for i in stations if i.name in stations_with_data]
```

There's also some nice markdown text (e.g. Background and DEA Use Case) in the old one that could be good to combine with the new one - maybe it would be worth going through both, and making sure any useful content from the old one is transferred across?

robbibt:
Something like this maybe: 

```
# Get list of stream gauges from BOM
stations = dea_bom.get_stations()

# Filter list to gauges with available data
stations_with_data = pickle.load(open(str('../Supplementary_data/Inundation_mapping/stations_with_data.pkl'), 'rb'))
stations = [i for i in stations if i.name in stations_with_data]

# Select a stream gauge
gauge_data, station = dea_bom.ui_select_station(stations)
```

MatthewJA:
@robbibt Could we replace the pickle with a JSON file? Or a numpy dump? Or maybe a CSV? I'm not sure what the actual data structure being pickled here is so I can't make a good concrete recommendation.

robbibt:
I think so, I'm pretty sure '../Supplementary_data/Inundation_mapping/stations_with_data.pkl' is just a table or list of IDs

robbibt:
@CEKrause If you tag me when you've made any other changes, I'm happy to replace the pickle and get the stations with data thing up and  running again

Edit: whoops wrong Claire (sorry Claire Martinez!)

CEKrause:
<p>We possibly could? But for the purposes of this example notebook, it's probably not needed.</p>

---

 <a href='https://app.reviewnb.com/GeoscienceAustralia/dea-notebooks/pull/688/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

CEKrause:
So it looks like the recent BoM web service update has gotten rid of stream gauges with missing data? I have been clicking around over the last couple of days and haven't found any, so I am happy to leave the notebook as is. I don't think we need to pickle anymore @robbibt.

robbibt:
@CEKrause That's a way better solution, yay! Should make it much nicer to use for other applications too

robbibt:
Before you merge, can you remove everything in here? ` '../Supplementary_data/Inundation_mapping/` I don't think we need any of it any more.

I also have a slight preference for the older, simpler notebook name "Inundation_mapping" over the longer one here to match most of the other real-world notebooks that have two or three work long titles, but totally your call! Everything else looks great to me

image

CEKrause:
I opted to keep the longer name, since it makes it clearer what the notebook actually does. I didn't realise what `inundation_mapping` was until it was pointed out. I have now incorporated all the suggestions, so I think this PR is good to go.

robbibt:
Last one was to get rid of the folder and contents here: `https://github.com/GeoscienceAustralia/dea-notebooks/tree/MDBA_notebooks/Supplementary_data/Inundation_mapping`

CEKrause:
Good catch. Done now.

robbibt (approved):
Looks great, a nice improvement for this functionality! The final thing that occurred to me is that I know the previous version of this notebook was used by a few external users... it could be worth posting on dea notebooks slack to let people know it's been updated/renamed in case they go looking for it and can't find the original

MatthewJA (approved):

PR #690, opened 2020-10-30
Fixed paths broken in #673

MatthewJA:
#673 renamed dea_temporaltools to dea_temporal; this PR fixes the two notebooks that this change broke.


robbibt (approved):
Good catch!

PR #694, opened 2020-11-02
Added waterbodies script and usage notebook

MatthewJA:
### Proposed changes
Added `dea_waterbodies.py` which queries the WFS server as well as a notebook demonstrating how to use it.


robbibt:
This is really, really nice! I really like how the script functions break the tasks into small chunks and how the API is similar between all of them. I've made some small changes to markdown to add some extra context to some of the cells, and rearranged the last bit slightly to discuss loading data for a waterbody first, then the slightly more advanced functionality of loading time series.

A few thoughts/suggestions:
- Geohashes are a pretty important part of using DEA Waterbodies data, but most users will be unfamiliar with them. I think it would be nice to give a bit more info between  `Getting data` and `List geohashes and polygons in an area` with a brief overview of what a geohash is, and how they are used for Waterbodies (e.g. as a unique ID with location info)
- One of the first things I wanted to do was use the list of geohashes loaded by `get_geohashes` to load data for all of those listed waterbodies. From what I can see, there isn't currently a way to do this, or if there is it's undocumented. Could `get_waterbodies` be modified to accept either a bbox _or_ list of geohashes as an input? (I guess the question then is whether we need a separate `get_waterbody` func at all?)

MatthewJA:
Thanks! :D

On your comments...

>  I've made some small changes to markdown to add some extra context to some of the cells, and rearranged the last bit slightly to discuss loading data for a waterbody first, then the slightly more advanced functionality of loading time series.

Thanks, this looks awesome.

> Geohashes are a pretty important part of using DEA Waterbodies data, but most users will be unfamiliar with them. I think it would be nice to give a bit more info between Getting data and List geohashes and polygons in an area with a brief overview of what a geohash is, and how they are used for Waterbodies (e.g. as a unique ID with location info)

Good idea. I wonder if it's relevant to discuss the fact that they're spatial IDs, or should we just state that they're unique identifiers? Depends how we want to balance information vs complexity.

> One of the first things I wanted to do was use the list of geohashes loaded by get_geohashes to load data for all of those listed waterbodies. From what I can see, there isn't currently a way to do this, or if there is it's undocumented. Could get_waterbodies be modified to accept either a bbox or list of geohashes as an input? (I guess the question then is whether we need a separate get_waterbody func at all?)

I originally had this actually... but because they're currently stored as CSVs, it's pretty messy to get them all (it's basically a loop over HTTP requests). On the dev sandbox we can do way better with Athena but that won't be a solution for the everyuser. But I could easily implement exactly that loop. Might even be able to parallelise it, but I'm not sure what the current "best solution" for parallel HTTP requests is in Python.

robbibt:
> Good idea. I wonder if it's relevant to discuss the fact that they're spatial IDs, or should we just state that they're unique identifiers? Depends how we want to balance information vs complexity.

I think mainly focusing on their use as a unique identifier, but touching on the spatial aspect very briefly (like, one sentence with a hyperlink to more info... "unique IDs that group nearby features together" etc) would be good. Doesn't need to be detailed, just enough that anyone can understand why the notebook doesn't just use something simpler like a basic numeric ID.

> I originally had this actually... but because they're currently stored as CSVs, it's pretty messy to get them all (it's basically a loop over HTTP requests). On the dev sandbox we can do way better with Athena but that won't be a solution for the everyuser. But I could easily implement exactly that loop. Might even be able to parallelise it, but I'm not sure what the current "best solution" for parallel HTTP requests is in Python.

If there's technical issues with loading multiple at a time I don't think it's critical for this PR, could always save for a future improvement if needed... just thinking from the perspective of making the whole API as consistent and user friendly as possible (it's already fine as it is).

MatthewJA:
Let's make it an issue and solve it in another PR.

MatthewJA:
Added a bit more detail on geohashes @robbibt, let me know what you think!

robbibt (approved):
Yep, looks perfect after the changes - just the right amount of extra detail

PR #705, opened 2020-11-05
Convert Scripts folder to `dea_tools` Python module

### LINKED ISSUE #538: Convert Scripts folder into package
We should convert `Scripts `into a separate package for:
* Simplified development process due to de-duplication with [deafrica-sandbox-notebooks](https://github.com/digitalearthafrica/deafrica-sandbox-notebooks)
* Testing
* Reduces path related import issues.
* Refocusing of `Dea-notebooks` and related repos to primarily hold just notebooks.

**Todo**:
* Decide if any functions / `.py` files aren't suitable for this transition
* Write `setup.py` etc
* Write tests
* Modify existing notebooks to import from package

MatthewJA:
This PR adds a `dea_tools` module to `Tools/`, which is installable with `pip`. It also updates _every_ notebook to use that module instead of inserting `Scripts/` into the path.

This is not meant to be a breaking change, so every notebook should still work. I ran every notebook that I could (a handful were too slow or too memory-hungry), so I have confirmed that they still work where possible.

The only breaking change that could happen is on the Sandbox, where user modules don't persist. The Sandbox will need to be updated to `pip install` this module from the most recent version (or cached version?) on load.

I don't want to merge this PR until we have resolved the Sandbox question (@Kirill888 may be the person to ask?), but I do want to kick-start the process of moving things over.

As discussed on Slack with @cbur24, once we have a functioning module for DEA, we can pretty easily build a DEAfrica component with a `dea_tools` dependency.

---

Here's what the old method of loading DEA scripts looked like:

```python
import sys
sys.path.insert(1, '../Scripts')  # or ../../Scripts etc

from dea_plotting import rgb
from dea_spatialtools import xr_rasterize
```

New method that I introduce here:

```python
from dea_tools.plotting import rgb
from dea_tools.spatial import xr_rasterize
```


robbibt:
This is definitely a good one to tag @Kirill888 (and maybe @whatnick and @omad ) as a reviewer too. 

As this is such a major change, my preference would be to run this in parallel and not merge it in to `develop` or `stable` for a little while, and probably have a chat with @caitlinadams and others from DEAfrica and Core about how this may align with standardising code and workflows between other non-DEA datacubes.

This is super cool though, and great progress on an issue we've been wanting to move forward on for a long time!

MatthewJA:
Good suggestions, and I also am open to having it in a parallel branch for a while - the target branch is currently develop, but we can change that.

That said, I actually think we're better pushing this into develop. Most people in DEA seem to be running their own parallel branch anyway, and adding another major branch (that touches every notebook) seems confusing at best and a nightmare to merge at worst. The longer it stays parallel to develop the worse that merge becomes, while right now it's pretty easy.

MatthewJA:
Updated the PR with a code example.

MatthewJA:
Oh BTW, if anyone wants to try this out, you can pip install it anywhere you like:

```bash
pip install git+https://github.com/GeoscienceAustralia/dea-notebooks@dea-tools
```

```python
import dea_tools.waterbodies
dea_tools.waterbodies.get_waterbody('r3dp1nxh8')
```

BexDunn:
My opinion is that we should go ahead and merge it, and make the changes, with future module developments with @caitlinadams  and our external collaborators to build upon this progress. We should announce the changes on the beginner's channel + potentially in a sandbox popup @whatnick ?

Kirill888:
there are some junk files included in the commit `.lock` an the like.

MatthewJA:
My bad, it was hard keeping track of this PR. I'll fix them up now

MatthewJA:
I think I got them all @Kirill888, thanks :)

cbur24:
Also we merged some changes to `classificationtools`, `spatialtools` today that should be pulled into this module

MatthewJA:
> Also we merged some changes to `classificationtools`, `spatialtools` today that should be pulled into this module

I'll pull 'em in now

MatthewJA:
Merged the changes, @cbur24

caitlinadams (commented):
Thanks for kickstarting the work on this @MatthewJA! 

I think it's fantastic to have gone through and not only have developed an initial module but to have updated all the notebooks to work with it as well.

From my perspective, I think this is a fantastic place to start from and serves as an excellent proof of concept of how we can be using python modules and packages to standardise how we use common functions. 

However, I have a few concerns, mostly related to investigating whether there's a way to do this for the Open Data Cube more broadly, rather than having a DEA specific module now, and then a DE Africa specific module that builds on it later. Instead, I'm curious if we can get to one module that works for both. This will save us a huge amount of effort in the long run and have much broader use outside GA at other organisations, such as CSIRO, AMA and Symbios.

I am also a little concerned about how this change might affect other users of the Sandbox, particularly those who've developed their own notebooks. Will the scripts folder remain? I worry that having both simultaneously will cause some potential confusion, particularly if the scripts folder effectively becomes deprecated. I'd like to see this addressed before this becomes part of the stable branch (and perhaps an outline of how it will be dealt with before merging into develop). If we are to remove it, we will need to engage with existing users to inform them of the change.

Thanks again for getting this started!

MatthewJA:
Thanks for the questions and insights @caitlinadams! We had a few discussions during Hacktoberfest@GA about how this might work between DEA and DEAfrica and the greater ODC community, so I've been thinking over those exact concerns for a little now.

Let me first stress I totally agree: we should have a general-purpose ODC tools library similar to the Scripts in dea-notebooks. I also think that we shouldn't make that yet. The gist of my thoughts is that cross-team communication is hard, balancing the different requirements of the codebase for each team is hard, and changing >2 teams worth of workflows at once is both hard and unnecessary because we can do it in steps. Implementing `dea-tools` in `dea-notebooks` changes nothing for non-DEA teams (except Sandbox users, mitigatable, see further down), doesn't make future development of a broader module any harder (in fact I think it's a necessary step regardless), and makes the code in `dea-notebooks` easier to engage with. Working on a broader implementation first would be slower and harder to integrate. I'm going to try and explain why I think this now!

> However, I have a few concerns, mostly related to investigating whether there's a way to do this for the Open Data Cube more broadly, rather than having a DEA specific module now, and then a DE Africa specific module that builds on it later. Instead, I'm curious if we can get to one module that works for both. This will save us a huge amount of effort in the long run and have much broader use outside GA at other organisations, such as CSIRO, AMA and Symbios.

I think a general ODC tools library that encapsulates this functionality is a great idea, and something we should 100% aim for. I see this PR as a step toward that: any process to move to an ODC module will basically have to run through the process behind this PR regardless (i.e. a transparent swap between the existing structure and the new structure). This is regardless of whether that repo is ODC-wide or DEA-wide, within dea-notebooks or somewhere else, and what it contains. Having a `dea-tools` module within `dea-notebooks` for now also lets us rapidly prototype in sync with our existing notebook workflows, without anyone here having to change their existing workflows very much (basically just editing modules instead of scripts when we need to add or change functionality). Trying to keep `dea-notebooks` in sync with a hypothetical broader ODC library would radically change our internal workflows and make it a lot harder to make our notebooks, and it also seems that it would be harder to integrate on the Sandbox (which would need to maintain version parity between the broader module and dea-notebooks). I would like to see us move to `dea-tools`, then move to a broader module, with Africa picking up the same utilities either before or after we do so.

There's a few different ways we can then go from `dea-tools` to some kind of general ODC tools:

- migrate `dea-tools` out of `dea-notebooks`, refactor, rename?
- build a new library that depends on `dea-tools`?
- build a new library which `dea-tools` depends on, transparently switch through to it (e.g. `from more-tools import *` in `dea_tools. __init__`)?

The third option is super appealing. This gives us immediate benefits from having a module, but provides a clear and user-friendly pathway to getting a broad set of tools in future. We can transition slowly to that, too, migrating functions out as they become mature in `dea-tools`.

It's also worth noting that _regardless_ of whether we implement a broader ODC tools (which I stress I'm in support of!) we will still need a DEA-specific module which uses that broader module as a dependency. The broader ODC module should not be specifically adapted to DEA or DEAfrica. We'd therefore need a module downstream from that. My proposal here is that we start with the downstream and migrate back upstream.

This is where I see DEAfrica fitting in, too. At the moment (as far as I know) DEAfrica maintains a sort of analogue to our Scripts folder - not quite a mirror because some things are different, but the gist is the same. If DEA switches to `dea-tools`, then `DEAfrica` can transparently swap out for `dea-tools` within their Scripts folder, or implement `deafrica-tools` which depends on `dea-tools`. We can quite easily sub out functions in `dea-tools` transparently with more general equivalents, and have `deafrica-tools` implement specifics atop them. Then in the future `dea-tools` would depend on the broader library, and we'd have a chain of three progressively more specific modules. Obviously this is not the only possible structure. For example, there's no reason to have `dea-tools` as the upstream rather than `deafrica-tools`. But there _is_ a good reason to not have `dea-tools` and `deafrica-tools` as independent dependencies on a broader ODC repo, because there's plenty of overlap between DEA and DEAfrica that would go unexploited in that scenario. I could also imagine a library which does both DEA and DEAfrica, but managing that (with all the quirks of Australian and African imagery plus our slightly different goals) sounds tricky. If we can keep development on each repo within the respective team, and contribute our changes back upstream, we'll be able to keep making rapid improvements and changes to our code without having to worry about whether these changes break workflows in other teams until we explicitly choose to (by merging upstream).

> I am also a little concerned about how this change might affect other users of the Sandbox, particularly those who've developed their own notebooks. Will the scripts folder remain? I worry that having both simultaneously will cause some potential confusion, particularly if the scripts folder effectively becomes deprecated. I'd like to see this addressed before this becomes part of the stable branch (and perhaps an outline of how it will be dealt with before merging into develop). If we are to remove it, we will need to engage with existing users to inform them of the change.

This is an excellent point. I think we need to maintain the Scripts folder, at least for now. There's simply no good way to have auxiliary code in a Jupyter notebook, which is a major limitation of the format. I would like to suggest that we aim to _not_ have auxiliary code outside a notebook as best we can, which reduces this issue. But other options exist; we can leave Scripts as is, which is no more confusing than the distinction between a Python "script" and a Python "module" (not to say that it isn't confusing - just that it isn't a unique issue). My favourite option would be to change the Supplementary Data folder to simply Supplementary, and then consider aux code as a supplementary artefact.

While I don't think local modifications to `dea_*` modules should be supported, we can absolutely make the change transparent if we choose to. This would involve (perhaps temporarily?) maintaining files in Scripts/ that are identical to the existing files from the outside (same location, same filename, same exports) but actually just refer to the `dea-tools` module. This is super easy - we just make files like `dea_spatialtools.py` which contain only `from dea_tools.spatial import *` and are otherwise empty (or perhaps toss out a `DeprecationWarning` when imported). This is then a fundamentally non-breaking change. I didn't do this in my PR, but I am happy to add it if that eases your concerns.

Let me know your thoughts!

caitlinadams:
Thanks for your thoughtful response, @MatthewJA! 

I think it's very sensible that you're looking at this as a first step in the process of developing a package that can be used across the ODC deployments. You raise excellent points around how doing too much too soon might make this process difficult and delay progress. 

While I agree with your thoughts around using this as an indicator of what the process might be like for other deployments, I would like to highlight that there are some challenges that would come from needing to build a module that works in all deployments. For me, the clear one is dealing with the different products in each deployment. I believe this is something we can potentially address with virtual products (based on discussions with @andrewdhicks and @uchchwhash), but this doesn't need to happen for a module that only serves DEA. 

My conclusion from this is that I think it's great for you to keep iterating and learning how to conduct this process through DEA, but that there still needs to be some effort into identifying whether a cross-deployment module is feasible, and some prototyping to prove we can solve the known blockers, and uncover any new ones that we haven't thought of. As such, I would recommend not polishing this work too much until some of these other factors have been established. I would like to work with you and interested parties to do this in a minimum-viable fashion, and I expect it could even enhance the DEA-specific development you're doing here. I would also suggest that we don't merge this until we're more confident about the path forward -- my thinking is that if we change the repo now, we'll then need to change it again in the future, which may lead to more confusion in the long run if the eventual ODC module ends up being quite different to this DEA iteration.

I agree that researching some solutions that make any potential transition from a DEA-focussed module to an ODC-focussed module easier is a good idea. In the long term, I would like to think that we could build an ODC-focussed module, which DEA then builds upon in its own module when more specific functionality is needed (I notice that you also think this). I am happy to support your suggested current approach of developing something DEA specific and then moving to a more general module, as long as you're comfortable with any rework that may eventuate when it comes to building a more specific DEA package based off a developed ODC package. While you say you would definitely need a DEA specific module, it is my hope that by bringing the community together, we can develop a general module that does by far the majority of the common tasks, with the DEA package becoming a smaller add-on for any specific tasks.

At this stage, I would recommend that we take some care when transitioning DE Africa to this sort of model, as we have just inducted a large number of new users, and we'll need to teach people how to use the new process (as well as update all our training material -- this stands for DEA too). As such, I think it would be a good idea to use DE Africa and DEA as a test case for validating whether it's possible and useful to build a shared ODC module. If it is, then we would progress to developing DE Africa from the ODC module, and DEA could progress with using the ODC module and determining what needs to be general and what needs to be specific.

It's good that you've started to think about how we could support the transitions, but I think we need some additional discussion from different perspectives to determine how best to proceed, particularly as there will be users outside GA that are impacted. As such, I would recommend keeping this on a separate branch (i.e. not merging to develop) until we have a better idea of the path forward. I want to make sure that the teams in both DEA and DE Africa have a clear understanding of what's being proposed, know what they can develop with, and how it might change in the future. I would hate for people to have a large amount of rework because they started using this module, only to have it change significantly with the development of a more general ODC module. I think we can get this clarity quickly by having some discussions and making a minimal prototype that indicates how the final module might work.

Let's keep talking to find a solution that enables great science!

MatthewJA:
Totally agree @caitlinadams! If I'm reading you right, we should work together to get to an MVP shared module, then implement DEA and DEAfrica modules atop this so that we can test whether a unified approach will work using DEA and DEAfrica as testbeds. That sounds OK to me. Let's meet at some point soon and figure out how we can push forward with a unified tools module? :)

caitlinadams:
Perfect summary, thanks @MatthewJA!

cbur24:
Just a thought, but this could be a great project for the ODC conference in February. Most of the stakeholders will be available then and a week-long hack means we should be able to make good progress

robbibt:
A reasonably small issue in the scheme of things, but if we do go for an all-of-ODC implementation, we should have a think about how this would interact with the [existing `odc-tools` package here](https://github.com/opendatacube/odc-tools), as it's kind of aimed at a similar type of code, i.e.

> "ODC features that DEA is experimenting with or prototyping with the intention of being integrated into odc-core in the future"

Kirill888:
`odc-tools` started off as prototypes for code to be maybe included in `datacube-core`, it have since evolved into "useful code that works with datacube library, but might be out of scope for datacube-core".  But `odc-tools` does not and should not contain any "installation specific" knowledge, like code to load NCI version of Landsat 8, for example. Even RGB code searchers for "red|green|blue" in the names of measurement and does not check for "nbar(t)_red" or things like that.

So to me there are several layers of generality/applicability

- `datacube-core` 
   - data agnostic (no hard-coded product names, band names etc)
   - should be useful for data processing tasks one might do with datacube
   - data visualization tools should probably not go there (they are hard to test and we want high test coverage in core, they also might depend on way too many externals libraries)
- `odc-tools`
   - data agnostic  (no hard-coded product names, band names etc)
   - Generic data visualisation tools with complex dependencies can be here
   - Generic metadata munging tools
   - Various prototypes that might end up in core
- `dea-tools`
  - Tools for working with specific datacube installations
  - hard-coded product/band names are ok, but best to make them "configurable" to help manage change and migrations across installations

MatthewJA:
From our discussion about this just now, I'm going to fix up this PR so that the Scripts folder passes through to the Tools folder.

MatthewJA:
Longer-term plan:

1. Make the tools module exist and work
2. Trial the module by manually installing it editable on Sandbox
3. Deploy the module on Sandbox once we are happy with it
4. Remove the Scripts folder
5. Generalise to de-tools + deafrica-tools

MatthewJA:
I'd like to force push a revert back to c0a82fc to avoid touching every notebook in this PR. Previously that was needed but now since we want to maintain the notebooks exactly as they are with a passthrough Scripts folder it's not necessary any more.

MatthewJA:
OK, I think that should work! This should now be completely compatible with existing notebooks and workflows. You should also be able to pip install dea-tools by running pip in editable mode from the dea-notebooks directory:

```bash
pip install -e Tools
```

MatthewJA:
I'll be trying out my own work for a bit using this branch, to see if anything has broken (the only thing I noticed was that the `__all__` definition in `dea_bom.py` was stopping my splat import, which I've now fixed by deleting `__all__`). I've run a few notebooks and they have all been fine so far.

CEKrause:
> I'll be trying out my own work for a bit using this branch, to see if anything has broken (the only thing I noticed was that the `__all__` definition in `dea_bom.py` was stopping my splat import, which I've now fixed by deleting `__all__`). I've run a few notebooks and they have all been fine so far.

The notebook that would be worth testing is the stream gauge wofs notebook in frequently used code. If that still works, then it should be ok?

MatthewJA:
Funnily enough that was one of the ones I tested! It worked fine as best I could tell, though it'd be great if you could double check

MatthewJA:
This one right? dea-notebooks/Real_world_examples/Mapping_inundation_using_stream_gauges.ipynb

caitlinadams (approved):
Hi @MatthewJA,

This is all looking good to me. For clarity, I would suggest the following changes:

- update the README in the Scripts folder to describe a little about how the Scripts folder now works (i.e. that these files now import the code stored in the Tools folder). It would also be good to explain that this ensures all current code works, but that users are encouraged to now use the DEA module from the Tools folder (see my next suggestion).
- update the README in the Tools folder to explain how Tools works, including an example of how to import the desired functionality in a notebook.
- update the README in the Tools folder to be a .rst file rather than .md file; this is mostly around if we ever build this into the documentation. It's not super critical, but it's easy to do now and will save someone having to do it in future. Feel free to ask for help with this if you haven't worked with .rst before.

These are mostly related to supporting users, so are nice to have rather than critical. I'm happy for you to decide if you want to make these changes and I can re-review if you want. Otherwise, way to go!

robbibt:
It might also be nice to include a standard comment at the top of all the re-factored scripts ([e.g. here](https://github.com/GeoscienceAustralia/dea-notebooks/blob/dea-tools/Scripts/dea_bom.py#L1)), explaining where the functions have moved to and why:

e.g. something like:

```
# Functions in this script have been moved to a new location to allow them to be imported into notebooks as a Python package: <link to new location containing scripts>
```

That way people can easily find them if they're looking for the source code

MatthewJA:
Should be all good to go @caitlinadams and @robbibt!

robbibt (approved):
This looks great to me, the extra documentation should help any existing users who may be confused about the changes!

Just flagging this as a future idea (not required for this PR): at some point could we move the remaining the notebookapp_* modules out of Scripts entirely and put them in Supplementary_data instead? That way we could just keep the `Scripts` around for compatibility in the short term, then remove it entirely later on once we have verified the new tools module works as intended.

MatthewJA:
Just gonna merge from `develop` into this first

PR #714, opened 2020-11-24
Added hard_leaves argument to fuzzywofs

MatthewJA:
Added a `hard_leaves` argument to FuzzyWOfS, which basically lets you assume that the marginals of the leaf nodes are pure. This means you'll get fuzzy results that are only about what leaf you ended up in, and don't account for the purity of those nodes - easier to mentally reconcile with the results you get if you just run WOfS normally.


robbibt (approved):
I haven't run this, but the changes look nice and contained so looks good for this to be merged!

PR #716, opened 2020-11-25
Added stream correlation notebook

MatthewJA:
Added a notebook that estimates the correlation between TCW in each pixel and nearby stream gauges. BOM Water Data Online is being super patchy right now so when I run it it only gives data for one of the two example stream gauges, but the notebook still works.


robbibt:
I'll have a closer look at this a little later on - for now, it would be good to update the introductory markdown to closer match the other real-world notebooks. For these ones we've tried to include the following three headers where possible, so that they can be used as part of DEA workshops and tutorials for beginners or non-technical users (`Frequently_used_code` notebooks are often a bit more relaxed and just include the single compulsory Description section):

1. Background: general overview of the ecological/environmental/management problem (not specific to DEA)
2. DEA/satellite data use case: how DEA and/or satellite data can help to address this problem
3. Description: A technical description of what exactly this notebook does

I think you have a lot of that content in there, just needs to be broken up a bit and elaborated on so that users can be taken on that journey of "yeah, that does sound like a tricky problem" > "I see how DEA/satellite data could potentially help solve that" > "I see exactly how we're going to solve that".

There's some nice examples here:

- https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Real_world_examples/Mining_rehabilitation.ipynb
- https://docs.dea.ga.gov.au/notebooks/Real_world_examples/Radar_water_detection.html
- https://docs.dea.ga.gov.au/notebooks/Real_world_examples/Coastal_erosion.html


I'd also replace the current "Stream gauge/pixel correlations" title with something a bit more descriptive like "Calculating pixel correlations with stream gauge data" or similar

robbibt:
Also possibly worth drawing your attention to two other ancient notebooks that deal with similar workflows, but were never migrated from the `master_archive` branch:

* Per pixel regression of tide heights: https://github.com/GeoscienceAustralia/dea-notebooks/blob/master_archive/05_Temporal_analysis/PerPixelRegression.ipynb
* Per pixel regression for groundwater: https://github.com/GeoscienceAustralia/dea-notebooks/blob/master_archive/09_Workflows/LinearModels-NDVI_recession.ipynb

Both of them were based on the [`lag_linregress_3d` function](https://stackoverflow.com/a/52711081/2510900) that calculates lots of per-pixel regression params in a nice and quick vectorised way (have always wanted to migrate a version of that func over to `Scripts`). Might be some things that could be harvested and integrated here?

MatthewJA:
I'll port that `lag_linregress_3d` function over!

MatthewJA:
I may also have autoformatted `dea_temporal`...

robbibt (changes requested):
Hey @MatthewJA , looks like there's a new conflict here with `dea_temporal.py`, might have been caused by some of the DEA tools updates?

There's also an error in the current version of the notebook that would be good to fix before we merge:
image

MatthewJA:
Did you get that error running it just now? If so, that's the BOM API failing.

I can fix the temporal issue though, will do shortly :)

robbibt:
That error is in the currently committed version of the notebook; just want to get a clean run stored in git so the error doesn't get rendered in the docs!

e.g. here: https://github.com/GeoscienceAustralia/dea-notebooks/blob/MatthewJA-streams/Real_world_examples/Wetness_stream_gauge_correlations.ipynb

MatthewJA:
Right! my bad

MatthewJA:
Look OK now @robbibt?

robbibt (approved):
Have made some updates to condense some of the notebook cells so it's easier to follow, and hopefully improved the run time a little using dask (it still takes a very long time to load the data, so in the future it might be good to update this to focus on a smaller area or time period). Good to merge!

MatthewJA:
Thanks for the modifications @robbibt!

PR #718, opened 2020-11-25
Added seasonality notebook

MatthewJA:
This PR adds a notebook that shows a few ways to detect and visualise seasonality in time series data. I've asked myself several times "is this time series seasonal?" or "has my deseasonalisation worked?" and this is a set of useful tools for helping to understand that.


robbibt:
This looks really great - nice and clear code and great documentation that nicely walks users through the steps. Some minor comments:
- Can you replace the link in the hyperlinked product name at the top to CMI instead of the website (e.g. https://cmi.ga.gov.au/data-products/dea/456/waterboards)? It's a more useful landing page for users that want more info I think (DEA Waterbodies is an anomaly in that there's no Explorer page; in time we may want to update all of the hyperlinked Explorer links to point to CMI too...)
- Please add this notebook to the README.rst file so it will render on the User Guide

A more general question... how do these workflows go when applied directly to irregularly spaced timestamps such as satellite observations rather than evenly spaced weekly data? Just thinking that many users may want to detrend raw data loaded directly from the datacube rather than apply any temporal interpolation, and it would be nice to give them tools to do so.

MatthewJA:
Detrending on uneven data is generally not supported. My approach to this is to upsample the data to weekly and interpolate over it, which should give you good results - then you can always reindex the seasonal trend onto the original time index and subtract it off that way. This particular notebook will work fine on unevenly sampled data because I implement this interpolation step, and there's an analogous function in xarray that'll do the same thing.

robbibt:
Yep OK, no worries - I wonder if it's worth including a note to that effect in the notebook near the section that you do the interpolation?

MatthewJA:
Added to the readme, changed the URL, added a note about interpolation

robbibt (approved):
Looks great to me!

PR #719, opened 2020-11-25
Trigger docs builds from notebooks repo

omad:
### Proposed changes
When changes are pushed to the `stable` branch, automatically trigger a re-publish of DEA Docs in the [dea-docs repository](https://github.com/GeoscienceAustralia/dea-docs/)

### DIFF
--- a/.github/workflows/trigger_docs.yaml
+++ b/.github/workflows/trigger_docs.yaml
@@ -1,0 +1,18 @@
+
+name: Trigger DEA Docs Build
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


robbibt:
This looks brilliant! I'm not an expert at this stuff but the workflow described above is exactly what we have been needing, so happy to merge this in. Will it need to be merged into `stable` too before it becomes "active"? 

Once this is merged, we can make another minor change to the repo and merge that into `stable` and see if it all updates correctly!

robbibt (approved):

PR #737, opened 2020-12-08
Added elevation notebook

MatthewJA:
A demo of estimating elevations using WOfS.


CEKrause:
- `stations_in_range` - does 'range' mean 'bounding box'? I couldn't spot a buffer, so I assume you're just looking for stream gauges within the specific bounding box? Could you please clarify this?
- is the `get_stream_data` function here the same as the one in the Scripts/BoM file? If so, can you just import that one? If not can you please add a sentence or two describing what you're doing here that's different?
- 'find the first shared observation' - shared with what?
- 'Mask the data to only include wet or not wet, and set everything else to nan.' - just checking there aren't more bit flags to include, like low solar angle etc? Ness had a list of a few she was using for waterbodies. Possibly for this application you want to err on the side of confidence though and exclude those other flags?
- 'Plot WOfS against stream gauge' - can you please add axis labels to the plot with the 0 and 128? I am not entirely sure what the plot is showing.
- 'We can then visualise this as a heightmap:' holy crap! This is amazing!!!!!!!!
- 'depth map' - I am not sure what you mean by 'Depth from maximum WOfS extent (m)'. You mean depth of water if the maximum amount of area is observed as wet? Can you convert to depth relative to the stream channel? So zero would be the top of the channel, and the floodplain would then be positive, and the stream negative? It may be too complicated?
- 'We can then apply the calibration to the WOfS summary:' - can you please add a label to the colourbar for the red/blue plot in this section? What is the anomaly?

Also, this notebook is absolutely nuts!!!! Well done.

MatthewJA:
image
Added a little widget to help investigate parts of the correlation plot.

MatthewJA:
> stations_in_range - does 'range' mean 'bounding box'? I couldn't spot a buffer, so I assume you're just looking for stream gauges within the specific bounding box? Could you please clarify this?

Made it explicit that it's a bbox.

> is the get_stream_data function here the same as the one in the Scripts/BoM file? If so, can you just import that one? If not can you please add a sentence or two describing what you're doing here that's different?

A bit different. I've added a sentence explaining that.

> 'find the first shared observation' - shared with what?

Clarified.

> 'Mask the data to only include wet or not wet, and set everything else to nan.' - just checking there aren't more bit flags to include, like low solar angle etc? Ness had a list of a few she was using for waterbodies. Possibly for this application you want to err on the side of confidence though and exclude those other flags?

There are more flags, but as you've said, let's err on the side of confidence.

> 'Plot WOfS against stream gauge' - can you please add axis labels to the plot with the 0 and 128? I am not entirely sure what the plot is showing.

Yep, done.

> 'We can then visualise this as a heightmap:' holy crap! This is amazing!!!!!!!!

:D Thank you!

> 'depth map' - I am not sure what you mean by 'Depth from maximum WOfS extent (m)'. You mean depth of water if the maximum amount of area is observed as wet? Can you convert to depth relative to the stream channel? So zero would be the top of the channel, and the floodplain would then be positive, and the stream negative? It may be too complicated?

Deleted this plot, since it's not that useful IMO. It's hard to estimate the top of the channel since it's not clearly defined, so I haven't done that conversion.

> 'We can then apply the calibration to the WOfS summary:' - can you please add a label to the colourbar for the red/blue plot in this section? What is the anomaly?

I can't seem to find the docs for colourbar labels for xarray plots... but really this plot is just kind of confusing, so I've deleted it.

> Also, this notebook is absolutely nuts!!!! Well done.

Thank you! :)

CEKrause:
Looks amazing! Is there a trick for getting your widgets to save their state? All your awesome plots are missing from the notebook output cells now that you've made them a widget. Otherwise, this one is good to go.

MatthewJA:
I think I can do something like that, I'll investigate!

MatthewJA:
@CEKrause I've added most of the plots back in by forcing the GUI back into inline mode :) the interactive plot won't show up still, but that's not a huge loss (it's more fun to poke around anyway)

CEKrause (approved):
Looks excellent!

PR #743, opened 2020-12-16
tools not implemented yet, adding a backup in the meantime

BexDunn:
I think I accepted the pr too early, as I'm not sure dea-tools is working yet? can you please check this out and see if I've fixed this or made it worse...
I changed the import stuff for dea-dask to check to see if the tools was working, and if not to grab it from scripts. 

I think the colors fix should still be ok, but if not please let me know...


BexDunn:
@emmaai this is probably ok to approve if you want to bring this older fix in?

PR #746, opened 2020-12-17
Folium map

Gabzgit:
### Proposed changes
Instead of having one classified map and one rgb map side-by-side, overlay them in folium.
The rgb is still not displaying correctly.


s-m-t-c:
Currently the folium map designed to replace the two image plot is currently displaying in solid black and white over the basemap. The goal is to have a single colour for the value 1 and transparency for the value 0 for the classified data over the RGB image.

Note that is considerably slower to load the folium map (8 seconds) compared to the images (1.5 seconds). This is probably irritating enough that I would consider sticking with the image method if it remains that slow but it would be neat to get the folium overlay working for the sake of it.

PR #752, opened 2021-02-05
Added WOfS plotting function

MatthewJA:
This PR adds a function to plot WOfS DataArrays in a way that matches the Alchemist thumbnail output. I've also used the function in the existing WOfS bitmasking notebook to demo it.


robbibt:
This looks great! Just a few questions - these don't necessarily need to be fixed right now!

- As "WOfS" is soon to be an outdated name, should we replace that terminology with "WO"/"Water Observations" in this function? e.g. rename the function to something like `plot_wo` instead?
- As it is currently, this function is for plotting the individual WO scenes themselves (i.e. what we used to call "WOFLs"). I can see people also wanting to plot the WO frequency maps with the official colour scheme... should we in the future somehow incorporate that functionality into this function, or have multiple funcs? For the second option, we might need to choose clear names for each so users know which one to use. 
- In the notebook, any reason for importing the func under a different name? `wofs as plot_wofs`? We might as well just choose a good name and use that directly

MatthewJA:
Done for 1 and 3; 2 I think makes sense but not for this PR. I've clarified the wording in the docstring to make it clear this is for bit flags.

robbibt (approved):
Looks great, thanks for the updates

PR #753, opened 2021-02-08
plot_wo bugfixes

MatthewJA:
Bugfixes in `plot_wo`:

- fixed an off-by-one error that caused nodata to appear as dry and terrain to appear as nodata
- moved from pcolormesh to imshow when possible to speed up rendering *and* make rendering work at all for large files


MatthewJA:
FYI: This PR looks more intense than it actually is because it's on the same branch as #752, it's really only like a 4 line change.

MatthewJA:
Also I am happy to do the "conflict resolution" :)

robbibt (approved):
Looks good to merge!

Issue #757, opened 2021-02-12
Update Subpixel waterbodies notebook to use geoserver DEA Waterbody data


robbibt:
Currently the Subpixel waterbodies notebook fails if the user does not read it carefully and download waterbody data into their sandbox instance. This is a little clunky and can be improved now that we have DEA Waterbodies data available via Geoserver with no downloading required.

![image](https://user-images.githubusercontent.com/17680388/107731735-cda3be80-6d4a-11eb-8b0a-be5f82985f0f.png)

robbibt:
I think this can be closed now

PR #758, opened 2021-02-14
Updated subpixel waterbodies to use geoserver, resolving #757

MatthewJA:


robbibt (approved):
Looks great, love how easy and simple the `get_waterbody` API is!

PR #763, opened 2021-03-03
Fix double fmask load bug when using `load_ard`

robbibt:
### Proposed changes
For a long time, `load_ard` has shown a curious behavior where the function appears to load Fmask cloud masking data twice when filtering by `min_gooddata` (i.e. the two discrete yellow sections below on the Dask status page below). This can add a large amount of time to all-of-time data loads, but hasn't ever been investigated in detail as Dask is a bit of a black box. 

old

It turns out that this is actually a super simple bug with major implications for load time. Essentially, [Line 365](https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Tools/dea_tools/datahandling.py#L365) creates a boolean that indicates whether each timestep was too cloudy or not to load data for (based on `min_gooddata`). When using dask, this is not evaluated until [Lines 368-369](https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Tools/dea_tools/datahandling.py#L368-L369). However, because each line is run independently, the boolean is evaluated separately each time, rather than being loaded only once and re-used.

The super simple solution is to simply `.compute()` Line 365 so it's evaluated only once, then used for both Line 368 and 369. This successfully removes an entire evaluation of the cloud mask layer:

image

A quick test on a tiny full stack of Landsat 5, 7 and 8 with `min_gooddata=0.99` cut total load time from 3 minutes 15 seconds to 1 minute 30 seconds. 🤯

### DIFF
--- a/Tools/dea_tools/datahandling.py
+++ b/Tools/dea_tools/datahandling.py
@@ -31,7 +31,7 @@ Functions included:
     last
     first
 
-Last modified: Jan 2021
+Last modified: March 2021
 
 '''
 
@@ -362,7 +362,7 @@ def load_ard(dc,
         print('Counting good quality pixels for each time step')
         data_perc = (pq_mask.sum(axis=[1, 2], dtype='int32') /
                      (pq_mask.shape[1] * pq_mask.shape[2]))
-        keep = data_perc >= min_gooddata
+        keep = (data_perc >= min_gooddata).compute()
 
         # Filter by `min_gooddata` to drop low quality observations
         total_obs = len(ds.time)
--- a/Tools/dea_tools/datahandling.py
+++ b/Tools/dea_tools/datahandling.py
@@ -362,7 +362,7 @@ def load_ard(dc,
         print('Counting good quality pixels for each time step')
         data_perc = (pq_mask.sum(axis=[1, 2], dtype='int32') /
                      (pq_mask.shape[1] * pq_mask.shape[2]))
-        keep = (data_perc >= min_gooddata).compute()
+        keep = (data_perc >= min_gooddata).persist()
 
         # Filter by `min_gooddata` to drop low quality observations
         total_obs = len(ds.time)


cbur24:
@robbibt good catch! I haven't tested this, but instead of `.compute()`, this might be a good candidate for the `.persist()` method.  To my reading of the docs, this should keep each chunk of `keep` loaded into distributed RAM rather than computing the entire object as a single future.

Couple of resources:
https://distributed.dask.org/en/latest/memory.html#persisting-collections
https://docs.dask.org/en/latest/dataframe-best-practices.html#persist-intelligently

BexDunn (changes requested):
I agree with @cbur24 - I think persist would be better than compute here "The operations client.persist(df) and client.compute(df) are asynchronous and so differ from the traditional df.compute() method or dask.compute function, which blocks until a result is available. The .compute() method does not persist any data on the cluster. The .compute() method also brings the entire result back to the local machine, so it is unwise to use it on large datasets. However, .compute() is very convenient for smaller results particularly because it does return concrete results in a way that most other tools expect.

Typically we use asynchronous methods like client.persist to set up large collections and then use df.compute() for fast analyses." https://distributed.dask.org/en/latest/memory.html#persisting-collections

robbibt:
So `.persist()` definitely does work, but to be honest after reading through all the docs I still don't think I understand them enough to make an informed decision between the two. In this case, it's a large, slow analysis with a tiny boolean result, which fits the criteria for *either* `.compute()` or `.persist()`. That said the boolean array is only used as an intermediate result, so `.compute()` is probably overkill as we don't need a standard array as an output?

Will do some further reading, and test if there's any performance differences!

robbibt:
This one really helped! Sounds like `.persist()` would indeed be a good tool for this job, thanks @cbur24 and @BexDunn! 
https://www.saturncloud.io/s/the-eager-data-scientists-guide-to-lazy-evaluation-with-dask/

cbur24 (approved):

BexDunn (approved):
I approve!

PR #769, opened 2021-03-12
Guided tutorial

BexDunn:
Adding the guided tutorial work into develop? Please check that the links work - I have checked but I have concerns based on the html build!

-otherwise, lets pull this in and then maybe into stable, and we can check again that the docs builds ok?


robbibt:
Hmm, just realised this notebook doesn't follow the DEA Notebooks template (e.g. no headers or footers). Should it? Apart from that I think everything looks fantastic!

For merging into `stable`, would it be OK if we did this on Monday rather than today? We could try and do it selectively and only merge in this subset of changes (not the broader `dea-tools` changes), but we would risk potentially getting the two branches out of sync (this happened during the refresh transition and it took a lot of attempts to fix up). On Monday we could send out a channel notification about the upgrade of `dea-tools` and how the new system works, then do the merge as soon as people have had a chance to see it

BexDunn:
Sounds like a plan. I'm out of juice on this today, add some headers and footers if you like! :)

alexgleith (approved):

erialC-P (approved):
Approved :)

PR #773, opened 2021-03-16
Scalable ML on the ODC

cbur24:
### Proposed changes
Porting over the Scalable Machine Learning notebook series from DE Africa.  During reviews, if you find typos and/or small changes that need doing, please push fixes back to the PR rather than simply commenting as I won't have time make every small change. 

As well as the notebooks, this PR also includes a new module called `dea_segmentation` that conducts parallel, tiled image segmentation.  There are supplementary figures also included.

The reason for two `readme` files is so the Sandbox/VDI user can have a rendered readme to guide them through the process (the read.md doesn't render in Jupyter Lab), and to acknowledge that some users might discover this series in the sandbox file browser rather than through github.

There are currently two sets of training labels in the `data/` folder: `crop_training.geojson` (2000+ polygons) and `crop_training_WA.geojson` (430 polygons).  The notebooks currently uses the WA dataset (which is a subset of `crop_training`) as it runs a lot faster. But the results are nicer using the whole dataset because the 'crop' labels are better distributed, and the user can then classify any region in Australia.  But with only two cpus on the default sandbox it takes >25 mins to gather the training data from the datacube (compared to 6 mins with the WA subset).    

I have not tested this on NCI as I am no longer set-up to run things since the Gadi transition


erialC-P (approved):
Really great set of notebooks with high value to the dea-notebooks repository

robbibt:
I'd love to go through these but don't have time this week... If we could leave this open until next week that would be fantastic!

cbur24:
> 
> 
> I'd love to go through these but don't have time this week... If we could leave this open until next week that would be fantastic!

Yes of course, @robbibt . Happy to leave it open until everyone is satisfied - there is a lot of material here so worth a thorough review process.

s-m-t-c (approved):
Read and ran everything through, excellent work. 

I changed some plot ordering / labeling for clarity.
Some  things such as the CPU calculation don't work on the NCI. 

```
ncpus=round(get_cpu_quota())
print('ncpus = '+str(ncpus))
```

cbur24:
@robbibt I'm planning on merging this one today as Sean and I would like to use these notebooks for a live demo next week. If you don't have to time to review today then is it fine if we merge them? When you get a chance to review we can make changes through a new PR?

robbibt:
@cbur24 I can review today! Will have a look now

robbibt (approved):
This is an incredible resource - amazing work! I want to go through this slowly in more detail for my own benefit later on, but for now I've just made some minor consistency updates (e.g. PEP8 formatting) and added the notebooks to the readme.rst so hopefully they show up in the user guide. Happy for this to be merged! 🚀

cbur24:
Thanks to everyone for the thorough reviews

PR #774, opened 2021-03-16
Update ARD_overpass_predictor.ipynb

Eric-git-999:
### Proposed changes
FIx start time for "timedummy" function, update from 2019 to 2020. This function orders dates, and needs to overlap the overpass predictions. Also Increase the number of iterations of "timedummy" from 300 to 500.


BexDunn:
Hey Eric! What does updating this do?

Eric-git-999:
Hey Bex, just a thing I overlooked. The "timedummy" is a list of dates, and I had them starting in 2019 (from when I first started working on this Notebook). 
If you are adding a start date, now in 2021, this may be later than the end of the "timedummy" if it starts in 2019.

I will look at a more elegant fix, something like grabbing the current month / year as the start point of the timedummy would be ideal, but in the mean time this update will keep it working properly :)

Eric-git-999:
Ok, fixed up properly now. "timedummy" grabs the current month and year :)

BexDunn (approved):
looks good

PR #779, opened 2021-03-25
Update stable with bugfix, workshop content, and kNDVI

MatthewJA:
Merging @alexgleith and @robbibt's changes into stable ahead of the ODC workshop.


robbibt (approved):
Probably don't need a review for this in the future - if you have the permissions feel free to just accept it yourself and then merge with the "Merge commit" option!

(all of the actual reviews have already taken place to get into `develop`)

MatthewJA:
@robbibt I think it's useful to have two people agree to merge things to stable, just adds an extra check in case we have to notify people again!

PR #784, opened 2021-04-07
Bugfix for plot_wo

MatthewJA:
Fixes large images in plot_wo. imshow seems to sometimes take add_colorbar and sometimes just colorbar... might be dependent on the version or something. At any rate this ducktypes for both options and has a fallback if they both fail.


BexDunn (approved):
looks useful

PR #788, opened 2021-04-21
Move TSmask notebook from DEA datasets folder to Scientific_workflows

robbibt:
### Proposed changes
The currently available TSmask data was uploaded as an experimental trial and is not currently a fully production-ready product. This has led to some confusion for users who expect it to be available for all timesteps and locations, or supported by data loading tools like `load_ard`.

Because TSmask is likely to be supplemented in the near future by advances to the method being developed by Peter Tan, ~~I propose temporarily removing this notebook from the main repo so that we do not give the impression that it is fully available as a production-ready product. We can easily restore this notebook from Git's history when the product is fully available, or alternatively move this notebook to its own branch in the meantime if we still want it available.~~

Updated to move notebook to Scientific_workflows instead


MatthewJA (approved):
Agreed. Feels like it should live on a branch until it's ready, or maybe within Scientific_workflows instead?

BexDunn (approved):
good idea

robbibt:
In response to feedback, have moved the notebook into the Scientific Workflows folder: 
https://github.com/GeoscienceAustralia/dea-notebooks/blob/remove_tsmask/Scientific_workflows/TSmask/TSmask.ipynb

BexDunn (approved):
better! :)

MatthewJA:
Big fan!

BexDunn:
I'm gonna squish, unless anyone says NOOOOO STOP?

robbibt:
YESSSSS SQUISH!

PR #798, opened 2021-05-20
Classificationtools geomedian fix

s-m-t-c:
### Proposed changes

* Changed the geomedian function, previous one was causing errors.
* Update to TQDM function also in this PR


s-m-t-c:
Still investigating if there is a problem with S2 loading in `collect_training_data`

cbur24 (approved):

PR #799, opened 2021-05-21
Upload Scripts/ dir to avoid huge git clones

omad:
A .tar.gz of Scripts/ will be uploaded to https://packages.dea.ga.gov.au/dea-notebooks/Scripts.tar.gz
on every change to the develop branch.

This doesn't add or modify any notebooks, just a GitHub actions script.


tebadi (approved):

MatthewJA:
Just noticed this. Why `Scripts` and not `Tools`, given that `Scripts` is a collection of files that almost entirely consist of imports from `Tools`?

@omad

robbibt:
More discussion here @MatthewJA: https://opendatacube.slack.com/archives/CGJV10LBH/p1621568981001700

tebadi:
In short, this step was redundant but we didn't know.  I'm going to delete this action.

PR #802, opened 2021-05-21
trimmed down collect training data to run exclusively off an input fu…

s-m-t-c:
### Proposed changes

The previous `collect_training_data` function grew to become unwieldy and tried to do a lot of work for the user behind the scenes. This made it difficult to debug, less flexibly and challenging to maintain the code base. By removing hard coded options this new function puts the onus on the user to define their own function and means it can be used across multiple datacube instances with different product names.

This is just a draft as I still need to change the other ML notebook, finesse some markdown text and remove unecessary imports.


cbur24 (approved):
@s-m-t-c I've made a number of minor refinements to the python functions, and cleaned up the markdown to align it better with this new paradigm. I recommend going over the changes I've made before merging. But otherwise, nice work!

PR #803, opened 2021-05-26
Classification tools xy error

s-m-t-c:
### Proposed changes
Small fix to classification.py changing docstrings and adding an extra error check to conform with the new paradigm.


cbur24:
I think the x,y error should also handle 'latitude' and 'longitude' keys as its possible to return arrays from the DC with spatial coords named like that

cbur24 (approved):

PR #822, opened 2021-06-09
changed func inputs to match new classificationtools

s-m-t-c:
Fixed small details with function to align with ctools updates.


cbur24:
Two things that I think need addressing:
1. Let's put a link to the `Scalable_ML` notebooks at the top of this notebook, so if a user wants to know more they have an idea of where to look. Make it clear that this notebook is a quick reference guide, while the other notebook series is an in-depth series.
2. In the `Extract training data using a shapefile` section of this notebook, can you add some comments about what is actually happening? Its an important part of the notebook but it doesn't have any markdown. It doesn't need to be extensive, but at least a short description of what each cell in that section is doing.

cbur24:
@s-m-t-c oh might also be worth bringing all the user parameters into the cell below `Analysis Parameters` heading.

cbur24 (approved):

PR #834, opened 2021-07-12
Replace Collection 2 FC and WO with Collection 3 FC and WO

robbibt:
### Proposed changes
Collection 3 is about to be officially announced this month, and support for Collection 2 is rapidly running out. To ensure that future users do not have to go through the pain of re-writing their code when Collection 2 is retired, we need to remove all Collection 2 code examples from this repo.

This PR updates notebooks that used `wofs_albers` and `ls*_fc_albers` to use the new Collection 3 layers instead (`ga_ls_fc_3` and `ga_ls_wo_3`). I have also made a minor clarification that the new Collection 3 layers require the use of `align` when native loading data.

Two issues remain:

* I've come to the [Subpixel waterbodies](https://docs.dea.ga.gov.au/notebooks/Real_world_examples/Subpixel_waterbodies.html) notebook and had a go at updating it to use `ga_ls_wo_3` instead of `wofs_albers`, but I ended up with some non-sensical results (I think because there isn't a perfect match up any more between `ga_ls_wo_3` and the Collection 2 all-time `wofs_summary`). @MatthewJA: any thoughts on the best way to adapt this notebook to use `ga_ls_wo_3` instead? We may want to consider taking it down temporarily until we can update it to C3 when the all-time summary is ready.
* There's still a bunch of Collection 2 Waterbodies notebooks in [Scientific_workflows](https://github.com/GeoscienceAustralia/dea-notebooks/tree/develop/Scientific_workflows/DEAWaterbodies) that use `wofs_albers`. These come up in the tag search in the user guide ([for example](https://docs.dea.ga.gov.au/notebooks/Scientific_workflows/DEAWaterbodies/DEAWaterbodiesToolkit/ErrorEstimationWOfS.html#index-4)), which means users may still potentially be using code based on C2 instead of C3. @MatthewJA @CEKrause: is there any chance we can get these relocated to the new Waterbodies repo before the Collection 3 announcement goes out?


MatthewJA:
@robbibt I ran it with C3 WO and the C2 WO summary with 25m pixels and nearest resampling - looks fine to me. A bit worse perhaps but that's to be expected. I'm not sure what the issue is that you've found?

robbibt:
I only got one or two data points in the final plots - I did resample to 30 m though, not 25 m. Any chance you could commit your version of the notebook? (or just send it to me on Slack?)

MatthewJA:
Yeah, I can do that!

MatthewJA:
I did try 30m but it just threw an error.

MatthewJA:
Updated @robbibt

Real_world_examples/Mining_rehabilitation.ipynb
```diff
@@ -9,8 +9,8 @@
     "* [**Sign up to the DEA Sandbox**](https://docs.dea.ga.gov.au/setup/sandbox.html) to run this notebook interactively from a browser\n",
```
MatthewJA, line 9: <p>Just WO, not WOs</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/GeoscienceAustralia/dea-notebooks/pull/834/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

MatthewJA (approved):
LGTM with the world's tiniest complaint.

Issue #806, opened 2021-05-26
Remove DEA Waterbodies from Scientific workflows and make DEA datasets notebook


CEKrause:
DEA Waterbodies has outgrown its life as a folder in `scientific_workflows`. Instead it should be highlighted with a notebook in `DEA_datasets` and the old code removed. 

Make sure to point to the new DEA Waterbodies repo.

* [x] Remove DEA Waterbodies from DEA Notebooks
* [X] Create `DEA_datasets` notebook for Waterbodies

robbibt:
We do have a `DEA_datasets` notebook for Waterbodies, so half of this issue is complete: https://docs.dea.ga.gov.au/notebooks/DEA_datasets/DEA_Waterbodies.html

@CEKrause or @MatthewJA: can I get one of you to do the Waterbodies clean-up in DEA Notebooks next week if you have time? (I'm currently removing all traces of Collection 2-based code from the repo, and at the moment lots of Waterbodies notebooks based on Collection 2 come up in the User Guide tag search)

robbibt:
Done!

PR #835, opened 2021-07-15
Move waterbodies to GeoscienceAustralia/dea-waterbodies

MatthewJA:
As requested in #834 and #806 this PR removes waterbodies code as it's all in dea-waterbodies now. A notebook in the datasets folder still needs to be added but should be done in a different PR.


robbibt:
Thanks so much @MatthewJA, this is brilliant! I think we already have a DEA Waterbodies notebook in DEA Datasets, so we should be good to close that issue! https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/DEA_datasets/DEA_Waterbodies.ipynb

robbibt (approved):
Looks good to me, thanks for this!

PR #837, opened 2021-07-16
Add new STAC metadata notebook

robbibt:
### Proposed changes
This PR adds a new notebook intended to provide a brief introduction to accessing and using DEA's STAC metadata:

1. - How to construct a STAC metadata API call
2. - How to search for STAC metadata and load the results into Python
3. - How to inspect and plot the unique STAC Items contained in the metadata
4. - How to inspect assets contained within a STAC Item
5. - How to download data using STAC metadata
6. - How to stream data into QGIS using STAC metadata (without downloading it first)

I'd love to hear any suggestions for improving this (either technical or regarding correct terminology), or any ideas for additional things to include. For example, in the future I'd love to add a section explaining how to iterate though large multi-page search outputs etc.


alexgleith:
I LOVE it @robbibt!

Only suggestion I have is that we could emphasise loading an XArray from the S3 URI instead of downloading first, but I think the use of QGIS is really cool too.

alexgleith (approved):

robbibt:
@alexgleith Yep, that would be a nice addition to the notebook, I agree! Will add something in

robbibt:
Hey @alexgleith , have added examples on streaming data to an `xarray` using `xr.open_rasterio`, and also how to use `from datacube.testutils.io import rio_slurp_xarray` to load data from the cloud and resample to match data loaded via `dc.load`

alexgleith (approved):
Love it!

Issue #849, opened 2021-08-11
Replace Scripts with Tools


MatthewJA:
Not using an actual install (which wouldn't work on the NCI), but instead:

```python
import sys
sys.path.insert(1, '../Scripts')
import dea_spatialtools
```
would become
```python
import sys
sys.path.insert(1, '../Tools')
import dea_tools.spatial
```

We also need to move auxiliary scripts to the extra data folder. We can then delete Scripts.

MatthewJA:
Ping @robbibt @BexDunn @s-m-t-c for visibility

robbibt:
This will be a reasonably large job, so we might want to divide labour a little (e.g. we could allocate `Beginners_guide`, `Frequenty_used_code`, `Real_world_examples` to different people etc

Should also note here that this is ideally a short term fix - in the longer run it would be much preferable to import directly from the installed module. We just need a solution for that on the NCI first while we still support that environment

robbibt:
Relevant to #731 and #732

MatthewJA:
Agreed on both

MatthewJA:
Now that we're on PyPI, maybe we can get dea-tools==latest installed on the Sandbox? That'd make this easier.

MatthewJA:
(whoops)

Issue #847, opened 2021-08-11
Use a GitHub action to mirror dea-tools in a separate repo


MatthewJA:
dea-notebooks is huuuuuge and it makes installing dea-tools a right hassle. Let's add a GitHub action that automatically mirrors Tools/ to its own repo.

MatthewJA:
Let's actually make this upload to PyPI instead.

MatthewJA:
Done! https://pypi.org/project/dea-tools/

PR #850, opened 2021-08-17
Added publish GH Action

MatthewJA:
This adds an action to publish dea-tools to PyPI when `stable` is updated.


MatthewJA:
Related #847 #849

s-m-t-c (approved):

PR #855, opened 2021-09-01
Added versioning to dea-tools

MatthewJA:
This adds versioning to dea-tools that hopefully doesn't break everything this time. :)

### DIFF
--- a/Tools/pyproject.toml
+++ b/Tools/pyproject.toml
@@ -1,3 +1,4 @@
 [build-system]
 # Minimum requirements for the build system to execute.
-requires = ["setuptools", "setuptools_scm"]
+requires = ["setuptools", "setuptools_scm", "wheel"]
+build-backend = "setuptools.build_meta"
--- a/Tools/pyproject.toml
+++ b/Tools/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+# Minimum requirements for the build system to execute.
+requires = ["setuptools", "setuptools_scm"]
--- a/Tools/pyproject.toml
+++ b/Tools/pyproject.toml
@@ -1,4 +1,4 @@
 [build-system]
 # Minimum requirements for the build system to execute.
-requires = ["setuptools", "setuptools_scm", "wheel"]
+requires = ["setuptools>=42", "setuptools_scm[toml]>=3.4", "wheel"]
 build-backend = "setuptools.build_meta"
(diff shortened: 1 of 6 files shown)


BexDunn (approved):
looks pretty good to me, but I can't 100% be sure nothing will break ;) I trust you though!

Tools/pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+# Minimum requirements for the build system to execute.
+requires = ["setuptools", "setuptools_scm"]
```
Kirill888:
```suggestion
requires = ["setuptools>=42", "setuptools_scm[toml]>=3.4"]
```
Copied from here:
https://github.com/opendatacube/datacube-core/blob/develop/pyproject.toml

In particular `[toml]` part is quite important

Tools/setup.py
```diff
@@ -130,10 +122,29 @@ def run(self):
...
+except LookupError:
+    try:
+        # no .git folder, so read from __version__.py
+        with open(VERSION_FILE_PATH) as f_version:
+            exec(f_version.read(), about)
+            version = about['version']
+    except FileNotFoundError:
+        # No __version__.py, yikes
+        raise RuntimeError('Build in-place with --use-feature=in-tree-build.')
+
```
Kirill888: You can just configure `setuptools_scm` in `pyproject.toml` like here

https://github.com/opendatacube/datacube-core/blob/dd46ceb7e027b1900547cf70696c72e4aa8ce552/pyproject.toml#L5-L6

```toml
[tool.setuptools_scm]
write_to="__version__.py"
root=".."
```

MatthewJA: Only if --use-feature=in-build-tree is enabled; this doesn't exist on Sandbox as pip is outdated. Otherwise pip copies the whole directory to /tmp and loses the git info

MatthewJA: Unless you just mean the configuration of setuptools_scm, in which case yeah that could be in the pyproject.toml

Tools/setup.py
```diff
@@ -130,10 +122,41 @@ def run(self):
...
 # Where the magic happens:
 setup(
     name=NAME,
-    version=about['__version__'],
+    version=version,
     description=DESCRIPTION,
     long_description=long_description,
     long_description_content_type='text/markdown',
```
Kirill888: yours is actually rst

Tools/pyproject.toml
MatthewJA: Thanks!

Tools/setup.py
MatthewJA: ah yep, good catch, thanks!

PR #865, opened 2021-09-08
Major update to remove Scripts, compatibility with new Sandbox image

### LINKED ISSUE #731: Relocate notebookapp scripts into supplementary data

### LINKED ISSUE #732: Update documentation to use `dea_tools` instead of the sys/scripts import
This amounts to changing the beginner's guide to using the new module.

### LINKED ISSUE #849: Replace Scripts with Tools

robbibt:
### Proposed changes
This update:
* Removes the `Scripts` directory, and updates all notebooks to load directly from `dea-tools` using:
```
import sys
sys.path.insert(1, '../Tools/')
from dea_tools.datahandling import load_ard
```

* Updates the repository for compatibility with recent updates to the DEA Sandbox image which are currently implemented on the [dev Sandbox](https://sandbox.dev.dea.ga.gov.au/).

* Adds a temporary fix to ensure that `dea-tools` modules can be imported locally: https://github.com/GeoscienceAustralia/dea-notebooks/blob/sep_update/Tools/dea_tools/__init__.py
```
try:
    from .__version__ import version as __version__
except:
    __version__ = None
```

### Closes issues (optional)
- Closes #731
- Closes #849
- Closes #732


MatthewJA:
I'd like to fix the `__version__` hack - basically the version is still totally gettable we just need to use setuptools_scm to do it. This isn't a blocker on getting the PR merged though; I'm happy to have it merged and then do a separate PR for that. Just noting my intent!

MatthewJA:
(not a review, just a note)

robbibt:
@MatthewJA Can you quickly check this one? Still looks like an error in the middle: https://github.com/GeoscienceAustralia/dea-notebooks/blob/sep_update/Real_world_examples/Subpixel_waterbodies.ipynb (Cell 12)
image

MatthewJA:
whoops, I fix

MatthewJA:
Try now @robbibt

Frequently_used_code/ARD_overpass_predictor.ipynb
```diff
@@ -60,8 +62,7 @@
    "source": [
     "import pandas as pd\n",
     "import numpy as np\n",
-    "import datetime\n",
-    "from datetime import timedelta"
+    "from datetime import datetime, timedelta"
```
MatthewJA, line 65: Is datetime used anywhere? I don't think it is, otherwise (as best I can tell) this would break.

Tools/dea_tools/__init__.py
```diff
@@ -1 +1,4 @@
-from .__version__ import version as __version__
\ No newline at end of file
+try:
+    from .__version__ import version as __version__
+except:
```
MatthewJA: Should be `except ImportError`; no bare `except`s!

MatthewJA (changes requested):
Got some minor requests.

robbibt:
Thanks @MatthewJA - the `datetime` thing was fine (`datetime` is a submodule of `datetime` and it's used here: `today = datetime.today()`). I've fixed the try/except!

MatthewJA (approved):
LGTM

PR #867, opened 2021-09-10
Update template and erosion notebooks to remove scripts

robbibt:
### Proposed changes
I accidently pushed directly to `develop` rather than raise a PR, so here's a PR for the `develop` > `stable` merge: updated the template notebook to use `dea-tools` instead of `Scripts`, and the same for the Coastal erosion notebook which was accidently missed in the first pass.


robbibt:
I realise in hindsight I probably should have run the dask-aware notebooks from a cluster with less cores to match what most people will have access to when they log in... hopefully it's OK to leave this as it is for now

image

BexDunn:
> I realise in hindsight I probably should have run the dask-aware notebooks from a cluster with less cores to match what most people will have access to when they log in... hopefully it's OK to leave this as it is for now
> 
> image

how many cores do most users have access to and what error will they get if they request too many?

robbibt:
@BexDunn They get two, and the number is automatically set by 
```
from dea_tools.dask import create_local_dask_cluster
create_local_dask_cluster()
```
We don't currently provide any instructions on how to set it higher, so I think the worst that can happen is that they run that function, see "2", and then think "I wonder why that's different from 15?".

robbibt:
I might just quickly re-run them with a smaller cluster after all - will push directly to `develop` again so it will be included in this PR

robbibt:
Done!

BexDunn (approved):
looks good

MatthewJA:
FWIW @robbibt Dask _should_ autoscale such that the number of cores doesn't really matter.

robbibt:
@MatthewJA Yep, this was more of an perception issue around external users who only have access to small clusters seeing a much higher number and questioning why they don't get access to that too... raises less questions in the long run by keeping the external stuff consistent!

PR #869, opened 2021-09-13
Remove ARD comparison code

robbibt:
As far as I can tell this isn't referenced anywhere on the repo, so simply removing these files should be fine.


MatthewJA (approved):
LGTM, we love a cleanup

PR #870, opened 2021-09-16
Notebook for exporting high quality satellite images from an interactive map

robbibt:
### Proposed changes
This notebook provides an interactive app that allows the user to select a region from a map, then export Digital Earth Australia satellite data as an image file. The function supports Sentinel-2 and Landsat data, creating True and False colour images, and  applying pansharpening to increase the resolution of Landsat 7 and 8 imagery.
        
Files are named to match the DEA Imagery and Animations folder naming convention so it will be easy for future DEA users to add new visualisations to those folders:
    
        "<product> - <YYYY-MM-DD> - <location> - <description>.png" 

Preview:

https://www.loom.com/share/e48952ef3ad0431db8747e3516cafe90


Supplementary_data/notebookapp_imageexport.py
```diff
@@ -0,0 +1,305 @@
...
+        else:
+            rgb_array = ds.isel(time=0).to_array().values
+
+        ############
+        # Plotting #
+        ############
+
+        # Create unique file name
+        site = reverse_geocode(coords=centre_coords)
+        fname = f"{sensor} - {date} - {site} - {style}, {load_params['resolution'][1]} m resolution.png"
```
BexDunn: Hey @robbibt ! can we please replace the spaces and commas in this filename template with underscores? e.g. f"{sensor} _{date}_{site}_{style}_{load_params['resolution'][1]}_m_resolution.png" . Site might also need string processing to replace any special characters e.g. " ',+,& " .

BexDunn (changes requested):
Hey ,please fix the filenames in the output files. Also looks nice otherwise! Great tool.

Supplementary_data/notebookapp_imageexport.py
robbibt: Hey @BexDunn! The current filename format is based on the DEA Images and Animations library naming convention here, so that visualisations produced by the tool can be easily added to those folders: https://geoscienceau.sharepoint.com/:w:/s/DEA/EWBOrEQPfnJOgk8hNveqFo0BT6KaoLbiuWIgiEolx_n29g?e=4MBZbf

It's been used there for a while and if we were to change it we'd need to go and re-name all the files in those folders as well... I do see the value in having a machine-readable option as well, so what if I added an option to the tool to set human-readable or machine readable file formats?

robbibt:
@BexDunn Made an update to give users the option to produce machine-learnable file names :)

BexDunn (approved):

PR #873, opened 2021-09-20
Add new functionality to Image export notebook and functions

robbibt:
### Proposed changes
This update makes the following changes to the `Export_satellite_images.ipynb` tool and code:

- allows the user to select the location of the image extraction independently from the export itself, allowing you to more easily customise the appearance of the output plot.
- sets the default image format to JPG, with the option to set it to PNG.
- adds a new `image_proc_funcs` param: an optional list containing functions that will be applied to the output image. This can include image processing functions such as increasing contrast, unsharp masking, saturation etc.  The function should take AND return a `numpy.ndarray` with shape [y, x, bands]. For example:
```
image_proc_funcs = [lambda x: skimage.filters.unsharp_mask(x, radius=5, amount=0.2)]
```


Real_world_examples/Exporting_satellite_images.ipynb
```diff
@@ -86,19 +86,21 @@
     "\n",
```
MatthewJA, line 86: <p>ma p -&gt; map</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/GeoscienceAustralia/dea-notebooks/pull/873/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

MatthewJA:
>allows the user to select the location of the image extraction independently from the export itself, allowing you to more easily customise the appearance of the output plot.

Maybe it's just too early in the morning for me but I couldn't quite figure out how to do this?

robbibt:
> > allows the user to select the location of the image extraction independently from the export itself, allowing you to more easily customise the appearance of the output plot.
> 
> Maybe it's just too early in the morning for me but I couldn't quite figure out how to do this?

All the map bit does now is produce a geopolygon. You run it once at the start:

image

And then you can run the "Export image" bit as many times as you want, customising the various parameters each time:

image

Does that make sense? The main change was to de-couple the map selection from the image export so you don't need to reselect the AOI every time you want to e.g. tweak the colours etc.

MatthewJA:
Oh yeah I somehow entirely missed the first cell there! LGTM

MatthewJA (approved):

PR #874, opened 2021-09-22
Update stable with develop

robbibt:
### Proposed changes
This update will add the new Export_imagery.ipynb notebook to the stable branch so that it will be available on the DEA Sandbox


MatthewJA:
Maybe we could get the tests on nbtests passing first, then merge that to develop, then merge that to stable? Just wary that the current version of develop has a failing test suite.

BexDunn (approved):
wow, this one is a big one

PR #876, opened 2021-09-24
Fix image export for Sentinel-2 NRT

robbibt:
### Proposed changes
Sentinel-2 NRT uses a different metadata format than Sentinel-2 Definitive, so the tool currently breaks with `satellites='Sentinel-2 NRT'`. This fixes this bug using a try/except to catch the different metadata fields.


MatthewJA (approved):
LGTM

Issue #877, opened 2021-09-24
TCW/B/G coefficients in calculate indices are for non-normalised values. 


BexDunn:
**Describe the bug/issue**
bandindices.py has normalise = True set on tasseled cap (Crist 85) coefficients and warns if normalise is set to false. Coefficients are for 0-10000 not for 0-1, so we should either remove the index from the normalisation, or if people want a normalised result then we will need the option to normalise after calculation. 

Preferred method of fixing: remove TCI from normalise list in function and/or add option to normalise after calculation. Not sure what to call that step.

BexDunn:
To clarify: I'm going to take the TCI out of the warning that says you're using it wrong if you don't normalise. The coefficients are set for values in the range 0-10000, if you normalise you will be using different thresholds and that is fine, but I usually use the non-normalised values and I don't want people to see a warning if they do this because using the non-normalised values is valid.

PR #878, opened 2021-09-24
Update warnings

BexDunn:
Remove warning where Tasselled cap indices for landsat are not normalised. This addresses issue #877

took the TCW, TCB and TCG out of the warning in bandindices.py

### Closes issues (optional)
- Closes Issue #877

### DIFF
--- a/Tools/dea_tools/bandindices.py
+++ b/Tools/dea_tools/bandindices.py
@@ -302,8 +302,8 @@ def calculate_indices(ds,
                               "refer to the function \ndocumentation for a full "
                               "list of valid options for `index` (e.g. 'NDVI')")
 
-        elif (index in ['WI', 'BAEI', 'AWEI_ns', 'AWEI_sh', 'TCW', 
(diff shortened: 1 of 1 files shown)


robbibt (approved):
Looks good!

robbibt:
Actually, should we remove the `'TCW_GSO', 'TCG_GSO', 'TCB_GSO'` ones too? Same thing should apply to them as well

BexDunn:
> Actually, should we remove the `'TCW_GSO', 'TCG_GSO', 'TCB_GSO'` ones too? Same thing should apply to them as well
I haven't gone and checked the original paper for that one - and didn't add the coefficients. Could go check? Not sure if it's 0-1 based or 0-10000 based without checking

robbibt:
I don't think it makes a difference - they should produce the same result whether it is given 10,000 or 1.0 scaled values

robbibt:
I don't think it makes a difference - they should produce an equivalent result whether it is given 10,000 or 1.0 scaled values

PR #884, opened 2021-10-15
Fix broken `tidal_tag` function in `dea_tools.coastal`

robbibt:
### Proposed changes
Recent updates to `xarray` broke code in the `tidal_tag` function (See: https://github.com/digitalearthafrica/deafrica-sandbox-notebooks/pull/299).

~This PR fixes the code, and also attempts to add the notebook to our tests so we can pick up broken code in the future.~

Update: decoupling the tests from this issue so we can get these bugs resolved quickly, then we can think about how to deal with accessing OTPS data for testing purposes later.


robbibt:
This bug correctly breaks all coastal notebooks in the repo, so it will be good to get this PR merged and we can sort out the tests later on.

BexDunn (approved):

PR #892, opened 2022-01-24
Add new interactive DEA Coastlines app and move all widgets to `dea_tools.app` submodule

robbibt:
### Proposed changes
This PR adds a significantly updated DEA Coastlines widget for interactively extracting shoreline data using coastline transects. This uses the fantastic new widgets framework put together by @caitlinadams  / DE Africa, which I can see becoming something we want to roll out to the rest of our interactive notebooks too.

At the moment, the interactive app includes:

- Selecting a transect from an interactive map
- Uploading multiple transects from GeoJSON or ESRI Shapefile
- Optionally exporting figures or output data
- Changing the map background

https://www.loom.com/share/7a54951e7ecb496e85f679ae16526be3

This PR also follow's DE Africa's lead by relocating interactive widget code to a new `dea_tools.app` submodule so we can more easily keep widget code in one central location (and remove it from `Supplementary_data` so that folder remains just for data, not code). I've updated the existing interactive notebooks to account for this change.


BexDunn (changes requested):
The transect extraction widget is incredible!!!! 😮 🥇 

- in future maybe split this PR into 2 PRs - changes to support the new notebook then the notebook itself, it's chonky 

 
Some thoughts on the notebook:


- colour map in statistics widget is struggling a bit around 0 - consider fixing this (not a major) I'm wondering if this is coming in from the WFS as I can't see where this is defined in the app or the script

- does it matter where you start drawing the transect from? I'm guessing yes, but I don't see any guidance on this. Dumb question.
- I'd like a comment under the resulting plot explaining 'along-transect distance' similar to that in the CMI, for dummies like me: e.g. "Negative values indicate that an annual shoreline was located inland of the baseline shoreline" or something to indicate the significance of what we see in the transects. I know the comment below the plot does say this, but I went to look up the CMI because my head didn't quite grasp the significance of the y axis at first go. 


Overall, this is TOTALLY FRICKING INCREDIBLE, and here are a few suggested changes :)

BexDunn:
This is awesome. I have made a couple of suggestions for you

robbibt:
@BexDunn Those are fantastic suggestions, thankyou! And definitely agree about splitting the PR in the future... I realised that about halfway through committing everything, but by then it was too late... 😱 

Will make the other suggested changes and let you know when it's ready!

robbibt:
Hey Bex, have updated the notebook!

> colour map in statistics widget is struggling a bit around 0 - consider fixing this (not a major) I'm wondering if this is coming in from the WFS as I can't see where this is defined in the app or the script

It's coming from Geopanda's new **`.explore` method that plots geodataframes on an interactive map. Unfortunately nothing I can do on my end - it's annoying!

> does it matter where you start drawing the transect from? I'm guessing yes, but I don't see any guidance on this. Dumb question.

Not a dumb question at all, it definitely matters which way you draw it. I've added some new guidance in the notebook and the tool:

image

image

> I'd like a comment under the resulting plot explaining 'along-transect distance' similar to that in the CMI, for dummies like me: e.g. "Negative values indicate that an annual shoreline was located inland of the baseline shoreline" or something to indicate the significance of what we see in the transects. I know the comment below the plot does say this, but I went to look up the CMI because my head didn't quite grasp the significance of the y axis at first go.

Have expanded the explanatory text below the plot:

image


Thankyou, these are great things to clarify!

PR #894, opened 2022-02-15
Updated image export notebook to an interactive app that requires no coding

robbibt:
### Proposed changes
This PR overhauls the current Image exporter app to make it use the new widget framework developed by DE Africa. This change makes the tool much easier to use, as all parameters are built into the app itself rather than needing to be specified by code. See demo here:

https://www.loom.com/share/71075ebc6bfe4d538618ae1c2dad5cb0

The tool supports the following functionality:
- Selecting date, satellite and imagery style
- Customising the colour stretch used to plot the image 
- Customising the output file format
- Customising spatial resolution to load data (in metres). By default, the tool will automatically set the best possible resolution depending on the satellites selected (i.e. 30 m for Landsat, 10 m for Sentinel-2). 
- Pansharpening: Whether to apply pansharpening (using the Brovey Transform) to increase the resolution of Landsat 7 and 8 imagery from 30 m to 15 m pixels. 
- Applying power transformations to rais imagery by a power to reduce bright features and enhance dark features. This can add extra definition over areas with extremely bright features like snow, beaches or salt pans.
- Applying unsharp masking to increase the sharpness of the output image. If activated, you can specify both the radius (in pixels) and the amount/strength of the sharpening.

Would love to get any feedback on usability, or edge cases that break the tool!


robbibt:
@cbur24 @alexgleith @caitlinadams Tagged you in case you eventually wanted to add something similar to DE Africa - no pressure to review if you're busy though!

CEKrause (approved):
This notebook is incredible! The only small tweak I'd suggest is adding a DPI adjustor to the Advanced menu so that you can export to a specific DPI if you need imagery for comms etc. 

This is amazing work.

robbibt:
@CEKrause Have added a DPI selector to the advanced menu - I ended up making it set the _minimum_ DPI for export, so if the image was smaller than the specified DPI it would be scaled up, but if the image was larger it would still be exported at its best resolution (even if that is higher than the specified DPI).

image

The reasoning for this was that I didn't want someone to accidently reduce the resolution of their data by specifying a lower DPI, but perhaps that's OK if that's what they request? Not sure, can be easily changed either way

PR #900, opened 2022-03-08
Nice colourbar landcover Animations

Kooie-cate:
### Proposed changes
Make nice colour bars for level 4 Land Cover Animations Possible
Previously they were a big mess. addition to the Landcover.py tools file makes them display correctly

also add an example to the land cover animations notebook of level 4 with colour bar

- [ X] Notebook created using the [DEA-notebooks template](https://github.com/GeoscienceAustralia/dea-notebooks/tree/develop)
- [X ] Remove any unused Python packages from `Load packages`
- [ X] Remove any unused/empty code cells
- [ X] Remove any guidance cells (e.g. `General advice`)
- [ X] Ensure that all code cells follow the [PEP8 standard](https://www.python.org/dev/peps/pep-0008/) for code. The `jupyterlab_code_formatter` tool can be used to format code cells to a consistent style: select each code cell, then click `Edit` and then one of the `Apply X Formatter` options (`YAPF` or `Black` are recommended).
- [X ] Clear all outputs, run notebook from start to finish, and save the notebook in the state where all cells have been sequentially evaluated


Kooie-cate:
Hi @robbibt @BexDunn @cbur24, @eloise-b  Belle wasn't able to finish reviewing this for me before she left so hoping someone can have a look over it for me? I first opened this request with a separate fiction to generate level 4 colourbars but Belle requested I combine it with the existing function. I initially published our landcover animation code with-out the ability to put a colour bar on animations using level 4 (our full classification) because it is very unwieldy. I just want to give a tiny over view so when you look at this you have some context, hope that's ok? 

the colour bars for animations have to be generated outside the  FuncAnimation() function or a new one is generated with every frame, there fore they need to be generated not using matplotlibs standard colour bar generation procedure because i's won't automatically space it and the level 4 colour bar is very wide due to all the text, this is why I actually do it on a figure with 2 sub plots, so one can be left empty and the colour bar can go over it.

Our colour scheme for level 4 colour bars are different to what we use for the actuall plots. why? becasue the full colour scheme has a number of classes we don't classify (yet?) and the numerical spacing between class numbers is uneven because it's categorical not continuous. Needed to make a different one for colour bars so the classes are even sizes. 

Please let me know if you have any questions about anything.

PR #903, opened 2022-03-16
Update animation notebook to use MP4s

robbibt:
### Proposed changes
I recently discovered the `from IPython.core.display import Video` function that allows us to preview video files inside a Jupyter Notebook. I've therefore made a minor change to the `Animated_timeseries` notebook to export MP4s instead of GIFs, which reduces the overall size of this notebook by around 50%, and hopefully encourage users to use the MP4 functionality instead as it is much faster (and produces much smaller output files).

There's a small chance this will also let videos render on the DEA User Guide which would be cool, but won't know until we test it!

(GIFs are still supported and the notebook includes an example to generate them - MP4s are just a better default)


CEKrause (approved):
This looks great. I spotted two small typos that I'll point out to you in person (easier than trying to explain them here). 

This notebook is looking really mature now.

PR #906, opened 2022-03-30
Add DEA training to repo

BexDunn:
Let me know what you think of the formatting.

I've added the ANU/DEA for Geospatial Analysts course material to the usage repo.


robbibt (approved):
Looks fantastic! Can't believe we didn't have this in there already!

PR #907, opened 2022-03-30
Update README.rst

BexDunn:
Added an explicit link in the readme to the usage page, and encouraged folk to add their work there.


robbibt (approved):
Great idea!

PR #908, opened 2022-03-30
Ah, add my conf paper

BexDunn:
I think my format is wrong. what format should I use?


robbibt (approved):
I don't think we have a set format, I've updated it for consistency!

PR #909, opened 2022-03-31
Remove spatial cross validation from Scalable ML

cbur24:
### Proposed changes
Something that I've been meaning to port from DE Africa for a long time; removing the spatial auto-correlation component from the cross validation methods in Scalable-ML. Spatial autocorrelation applies less often than I initially believed, so I've removed it from the default example as that should be more applicable to most use-cases.


robbibt (approved):
Looks good!

PR #910, opened 2022-03-31
Update phenology notebook and script

cbur24:
### Proposed changes
Simplify the `xr_phenology` function by removing the smoothing and interpolation functions, and placing these steps in the notebook instead,


robbibt (approved):
Looks great, I thought there might be some good things to update here following all the DEAfrica progress!

PR #912, opened 2022-04-01
Add morphological filtering to `load_ard`

cbur24:
### Proposed changes
Proposing a new feature for `load_ard`, the ability to apply binary morphological filters to the cloud mask.  `datahandling.py` is updated and the `Using_load_ard.ipynb` notebook.


robbibt (approved):
Fantastic! This will be particularly useful when the Sentinel-2 Collection Upgrade is done, because Fmask will be updated to match Landsat's, with no automatic buffer applied. Because of this, we might want to consider in the future whether to set a sensible default to this param rather than defaulting to `None`, as it might make more sense to mimic the Fmask defaults (e.g. with buffering built in) unless the user has a specific requirement for no buffering. But for now, the `None` default is fine!

PR #914, opened 2022-04-06
Conference contribution GeoPython

sebfoe:
Citation style after APA

### Proposed changes
Adding a conference contribution at GeoPython 2021 using the ODC

- [X] No notebook


robbibt (approved):
This is fantastic, thanks @sebfoe! I made a small change to match the formatting of the others (for some reason we seem to be using a bespoke citation format here, so later on we can edit all of these to use something more standard.

Issue #916, opened 2022-05-03
some notebooks not updated to collection 3


Kooie-cate:
A number of notebooks in the repo still load landsat collection 2 dataset. For example a number of the Beginners_Guide notebooks including 04_Loading_Data and 05_Plotting load collection 2 Geomedians.

robbibt:
Running this:
```
import glob
pattern = '*/*.ipynb'
query = '_geomedian_annual'

for filepath in glob.iglob(pattern, recursive=True):
    with open(filepath) as file:
        s = file.read()
        if (s.find(query) > -1):
            print(filepath)
```
Indicates these notebooks need to be updated:
```
Beginners_guide/04_Loading_data.ipynb
Beginners_guide/09_Parallel_processing_with_Dask.ipynb
Beginners_guide/05_Plotting.ipynb
Beginners_guide/03_Products_and_measurements.ipynb
Frequently_used_code/Exporting_NetCDFs.ipynb
Frequently_used_code/Machine_learning_with_ODC.ipynb
```

robbibt:
I think this has been addressed in #917 so closing for now - let's re-open if we notice any other notebooks that are using C2 data!

PR #917, opened 2022-05-20
Update notebooks to use C3 derivatives

robbibt:
### Proposed changes
This PR updates notebooks in the repo to use the newer Collection 3 geomedian and FC percentiles products.

The tests also need to be updated to point to the new products, but we can do that as a later PR.

### Closes issues (optional)
- Closes Issue #916


robbibt:
@cbur24 Tagged you as I made a bunch of changes in the ML notebooks, mainly to swap to the new data, but also some other things that were causing issues (e.g. there was a bit that dropped several columns off the training data to drop x/y data, but it was removing valid data so I got rid of it... everything seems to work OK though)

Kooie-cate (approved):
awesome, all the substitutions look good, notebooks all tested no problems found with any of them.

Kooie-cate:
this might be a question for @cbur24, in Scaleable_machine_learning/4_Classify_satelite_data` cell 16 throws this error repeatedly `UserWarning: Loky-backed parallel loops cannot be called in a multiprocessing, setting n_jobs=1
  n_jobs = min(effective_n_jobs(n_jobs), n_estimators)` 
I just wanted to make sure this wasn't a **new thing**, if this is a known thing that's just a result of using Dask with these modules, that's all good, It dosn't stop the notebook working. 

gosh sorry everyone I clicked the wrong button

erialC-P (approved):
Reviewed all and happy to approve.

robbibt:
Thanks @Kooie-cate and @erialC-P! I'll merge this PR, but Chad should still be able to comment here about the `Loky-backed parallel loops` issue

cbur24:
@robbibt @Kooie-cate 
The `Loky-backed parallel loops` issue happens because the model that's output in the previous notebook has the `n_jobs` parameter set as > 1, which doesn't gel with dask handling the parallelism.  To fix this open the model (cell `[5]`) with the following call:

`model = load(model_path).set_params(n_jobs=1)`

PR #919, opened 2022-05-23
Support for FES2014 tide model via CNES aviso-fes.

rtaib:
- Coded as replacement drop-in for coastal.py
- See documentation in pyfes_model.py

### Proposed changes
A drop-in alternative to the OTPS tide model in `Tools/dea_tools/coastal.py` which defaults to OTPS and then tries to use `aviso-fes`, if installed, to calculate tide heights. It is coded as a new `pyfes_model.py` with the same method signatures as the current code to allow seamless swapping without changes to the current code. However, the user needs to set the `FES2014_OCEAN_INI` environment variable as explained in the docstring for the module.
An example of use is provided in `Real_world_examples/Coastal_erosion.ipynb`.

### Closes issues (optional)
- NA


robbibt (approved):
This looks great to me, thanks for this! Depending on whether we can find an easy way to install aviso-fes, we might end up adopting a similar approach later on as the default way to model tides in the DEA Sandbox too

PR #927, opened 2022-06-23
New generating animations interactive notebook

robbibt:
### Proposed changes
This PR adds a new notebook `Real_world_examples/Generating_satellite_animations.ipynb` that allows a user to interactively draw a box on a map, and export a custom animation (with the ability to control colour stretch, rolling medians, image sharpening, resolution, cloud coverage).


robbibt:
Hey everyone, thanks so much for the feedback on this last time! I've just pushed an update that makes the following updates:

- Changed initial zoom to focus on just Sydney, and reduced the default time-period to 3 years to hopefully reduce memory crashes
- Added a new "Temporal resampling" option to allow you to create monthly, quarterly and annual timestep animations
- Added a new "DEA Coastlines overlay" option to plot annual shorelines over the top of the animation
- Fixed black frame bug that Claire hit with her big animation by setting nodata to `NaN`
- Added missing `groupby` to data load

Example of an annual animation with DEA Coastlines overlay for Brisbane!
https://user-images.githubusercontent.com/17680388/176623771-3559db1e-3a9f-464d-83b1-2487cf71ecfb.mp4

cbur24 (changes requested):
This is amazing, @robbibt. incredible how far the interactivity in these notebooks has come in the last year or so.
Couple of comments/issues:
- An issue I ran into, whenever I tried running animations with .mp4, the animations stop after a few frames, I tried this with both Landsat and Sentinel-2. With .gif it worked fine. Maybe it was just an issue with my video player - not sure

- I noticed that the entire time-series is brought into memory before resampling happens. It should be more memory efficient (and perhaps faster) if in the case of resampling you delay the `.compute()` method until the time-series is reduced (in which case maybe bring the reducing step into the `extract_data` function). 
- In line 187 of `animation.py` the dask client is closed before returning the computed time-series, is that correct?
- In my experience, `client.shutdown()` seems to work better than `client.close()` - I think that's what those task-pending warnings are.

robbibt:
@cbur24 Super handy feedback, will have a look at all of them and see what I can change. The animation stopping after a few frames is weird - might be player/codec related? If you try running it on your phone or even directly opening it in Chrome, does it work any better?

cbur24:
> @cbur24 Super handy feedback, will have a look at all of them and see what I can change. The animation stopping after a few frames is weird - might be player/codec related? If you try running it on your phone or even directly opening it in Chrome, does it work any better?

Ah yes you're right, the mp4s are playing fine in the browser but not on my computer - nevermind!

robbibt:
Hey @cbur24 , implemented all the updates above except the `client.shutdown()` one - for some reason it was causing some really strange errors when I tried to load new data after applying resampling - will have a look at this in more detail later on.

I need this notebook for a meeting in 30 mins so am going to merge this PR now, but super happy to revisit anything later on too! Thanks for the review - lots of great catches!

PR #928, opened 2022-06-24
Fix failed Dask worker pod launch due to invalid username characters

jmettes:
Since we've moved to using email addresses in Cognito, this code is now passing `@` symbols into the Kubernetes worker pod and causing it to fail to create.

> Failed to create pod in namespace sandbox - Pod "dask-worker-935744a16ab74944bff079717a6a5ba0-ls628" is invalid: metadata.labels: Invalid value: "jonathan.mettes@ga.gov.au": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')

Restrict `jupyterhub_user` parameter to alphanumeric characters.

### DIFF
--- a/Tools/dea_tools/dask.py
+++ b/Tools/dea_tools/dask.py
@@ -21,7 +21,7 @@ Functions included:
     create_local_dask_cluster
     create_dask_gateway_cluster
 
-Last modified: March 2020
+Last modified: June 2022
 
 '''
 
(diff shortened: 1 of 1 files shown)


robbibt (approved):
Awesome! I also hijacked this PR to add the code to automatically close any existing clusters

emmaai (approved):
tyty Jonathan!

PR #932, opened 2022-08-01
Waterbodies fix

vnewey:
Updated code to use DEA Waterbodies V2


BexDunn (approved):
hopefully I haven't broken anything in approving this one :)

robbibt:
At some point it would be great to add some tests to the DEA Waterbodies code - given it doesn't need datacube access to access the data it should be a bit more straightforward than some of our other notebooks!

Issue #926, opened 2022-06-22
rasterstats dependency missing in pip install


anton-seaice:
**Describe the bug/issue**
When dea-tools is installed in pip, rasterstats is not installed but is needed for dea_tools/spatial.py

**Steps to reproduce**
1. pip install dea-tools in a clean environment
2. In python, attempt to "import dea_tools.spatial"
3. ModuleNotFoundError is returned

**Expected behaviour**
Module loads without error.

**Environment information**
pip install of dea-tools

anton-seaice:
geopy module is also missing

PR #934, opened 2022-08-01
Add missing rasterstats and geopy dependency

robbibt:
### Proposed changes
This PR addresses #926 by adding both `rasterstats` and `geopy` to the DEA Tools `setup.py` file.

### Closes issues (optional)
- Closes Issue #926


CEKrause (approved):
I didn't run this, but I assume this doesn't cause conflicts?

robbibt:
It's really hard to test - we'd need to install it on a fresh environment outside of the DEA Sandbox or NCI... however, both of those packages are already installed and working fine on the Sandbox, so that makes me think they work together without conflicts. I might just tag the original bug reporter and see if they can test on their end once it goes live

PR #936, opened 2022-08-04
Add funcs from DEA Coastlines and DEA Intertidal to DEA Tools

robbibt:
### Proposed changes
As part of DEA Coastlines, we developed several functions that may be useful to users of DEA Tools as well. This functions also need to be called by the new [DEA Intertidal repo](https://github.com/GeoscienceAustralia/dea-intertidal), so it makes sense to move them to one central location, rather than duplicating the code in multiple places.

This includes the following new functions:

#### `dea_tools.coastal`
- `dea_tools.coastal.model_tides`: A replacement for the existing `tidal_tag` func that uses the open-source FES2014 tidal model. This won't currently work without users downloading some additional model files, but eventually the plan is to get those model files installed by default on the Sandbox (like we do with the current OTPS tidal model).
- `dea_tools.coastal.pixel_tides`: A more advanced tidal model function that models tides across the 2D extent of a satellite imagery dataset, then interpolates that data into each pixel of the dataset. Used in DEA Coastlines to create seamless tidal composites and shorelines (with no discontinuities along study area boundaries).

#### `dea_tools.datahandling`
- `dea_tools.datahandling.parallel_apply`: A function for applying a custom function to each timestep in an `xarray` in parallel, using multiprocessing from the `concurrent.futures` package (recommended by @omad)

#### `dea_tools.spatial`
- `dea_tools.spatial.hillshade`: A function for generating hillshades on DEM data, using custom sun angles and azimuths.
- `dea_tools.spatial.sun_angles`: A function for extracting average sun angles and azimuths from DEA Collection 3 metadata for each timestep in a satellite dataset. Can be used to identify pixels that are likely to be affected by terrain shadow in surface reflectance data (less of an issue in Australia with NBART, but might still be useful

### DIFF
--- a/Tools/dea_tools/coastal.py
+++ b/Tools/dea_tools/coastal.py
@@ -22,8 +22,10 @@ Functions included:
     tidal_stats
     transect_distances
     get_coastlines
+    model_tides
+    pixel_tides
 
-Last modified: January 2022
+Last modified: August 2022
 
 '''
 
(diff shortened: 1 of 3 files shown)


vnewey (approved):
Looks good!

PR #937, opened 2022-08-05
fix to Surface Area Duration NB

vnewey:
Updated to Surface Area Duration NB to work with a newer version of pandas. + updated for DEA Waterbodies V2

### Proposed changes
Updated to Surface Area Duration NB to work with a newer version of pandas. + updated for DEA Waterbodies V2

- [x ] Test notebook on both the `NCI` and `DEA Sandbox` (flag if not working as part of PR and ask for help to solve if needed)


robbibt (approved):
Hey @vnewey, looks great! I pushed a tiny update to fix two missing images, and change a few CMI links that pointed to the old Waterbodies CMI link.

One more that's not necessary for now but would be work doing would be to update this notebook to use the `get_waterbodies` and `get_timeseries` funcs from `dea_tools.waterbodies`. I don't think they existed when this was first written, but they're a much easier method than accessing the CSVs directly. Maybe something to create an issue for?

PR #940, opened 2022-08-08
Fix missing function import in `pixel_tides`

robbibt:
### Proposed changes
This PR fixes a missing import of the `parallel_apply` function from `dea_tools.datahandling`, which is currently needed for the `pixel_tides` function to work.

Once merged I'll have to release a quick new package version and do another `develop` to `stable` PR.


vnewey (approved):
Looks good!

PR #943, opened 2022-08-18
Update DEA Tools to support autodocs generation in the User Guide

robbibt:
### Proposed changes
The PR here (https://github.com/GeoscienceAustralia/dea-docs/pull/30) will enable us to render functions and modules from DEA Tools into the DEA User Guide as a read-the-docs API, e.g.:

image
image

This PR expands the current placeholder index file in DEA Tools to add additional modules to the read-the-docs, copying [a similar approach to DEA Africa here](https://github.com/digitalearthafrica/deafrica-sandbox-notebooks/blob/main/Tools/index.rst). This should add all our functions to the User Guide when it's merged.

Next steps after this might be to browse the automatically generated docs and check if there's any docstrings in DEA Tools that we want to fix up to improve how it's rendered in the User Guide!

### DIFF
--- a/Tools/dea_tools/bandindices.py
+++ b/Tools/dea_tools/bandindices.py
@@ -1,8 +1,8 @@
-## dea_bandindices.py
 '''
-Description: This file contains a set of python functions for computing
-remote sensing band indices on Digital Earth Australia data.
+Calculating remote sensing band indices (e.g. NDVI, NDWI)
+'''
 
+'''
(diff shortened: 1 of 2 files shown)


robbibt:
Have also done a test of splitting the DEA Tools module docstrings into two parts, so only the short summary gets rendered automatically into the `readthedocs` module table

omad (approved):
Awesome! :D :D

PR #946, opened 2022-08-26
Add clarifying notes and links to wiki, particularly for ML pages

treefern:


robbibt:
Hey @treefern - this is great, I guess my only concern though is that things will end up a bit hap-hazard if we dive in and add the wiki link to individual notebooks like the machine learning ones above - ideally we want to keep the format and structure of all the notebooks and readmes identical so they're as easy to use as possible. Maybe we could instead include a standard link to the Wiki on all of the main sections of the folder .RST indexes, similiar to the current "Citing DEA Notebooks" section we have on all of those pages?

That way there'd be less to maintain, and things can be kept nice and consistent across the repo.

https://github.com/GeoscienceAustralia/dea-notebooks/tree/develop/Beginners_guide#citing-dea-notebooks

PR #950, opened 2022-09-05
Add try to account for different param names in different versions of `pyTMD`

robbibt:
Recent changes to the `pyTMD` tide modelling function have changed the parameter format of certain functions from CAPITAL CASE to lower_case params.

### Proposed changes
This PR accounts for this by adding a try/except to the `model_tides` function.


vnewey (approved):
I haven't tested it, but it looks good. :)

PR #951, opened 2022-09-05
Move tide modelling funcs from `dea_tools.coastal` to function level instead of module level

robbibt:
### Proposed changes
The DE Africa Coastlines docker build is currently failing because it can't load the `dea_tools.coastal` module because of these imports:
```
try:
    from otps import TimePoint
    from otps import predict_tide
except ImportError:
    from dea_tools.pyfes_model import TimePoint, predict_tide
```
https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Tools/dea_tools/coastal.py#L48-L52

This is because `otps` is bespoke and not properly open source, so we comment it out in the `dea_tools` setup.py to avoid it being installed when non-Sandbox users try to install `dea_tools`: 
https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Tools/setup.py#L47

To get around this, I've moved this import section out of the top level of the module, and into the functions that actually use it (`tidal_tag` and `tidal_stats`). This should allow users to use the other functions inside the `dea_tools.coastal` module without getting an `Import error` when they first try and import the module itself.

Have tested this still works using the Tide_modelling.ipynb notebook 😃


vnewey (approved):
Nice solution Robbi.
Once again I haven't tested it, but it looks good.

PR #952, opened 2022-09-09
Move pyfes imports inside funcs in `pyfes_model.py`

robbibt:
### Proposed changes
This should prevent issues with the module failing to load due to `pyfes` not being installed (and not being included in the `seup.py`.


vnewey (approved):
Looks good

vnewey (approved):

PR #958, opened 2022-10-06
Update module and func docstrings to improve autodocs

robbibt:
### Proposed changes
This PR makes changes to the module and function docstrings in `dea-tools` to improve how they are rendered in the autodocs in the user guide here:
https://docs.dea.ga.gov.au/notebooks/Tools/index.html

Main changes include: 

- Removing superflous "Description: " into on the module docstrings. 
- Removed the "Available functions" bit as this info is now rendered in the User guide itself
- Fixed some broken function docstrings that were rending weirdly in the docs
- Improved the formatting of lists in func docstrings

One other small change I made was to add a deprecation warning to an ancient function I wrote ~5 years ago that is no longer a good way to do things - I doubt anyone uses it, but at least with a deprecation warning we can retire it in the future.

- [NA] Notebook created using the [DEA-notebooks template](https://github.com/GeoscienceAustralia/dea-notebooks/tree/develop)
- [NA] Remove any unused Python packages from `Load packages`
- [NA] Remove any unused/empty code cells
- [NA] Remove any guidance cells (e.g. `General advice`)
- [NA] Ensure that all code cells follow the [PEP8 standard](https://www.python.org/dev/peps/pep-0008/) for code. The `jupyterlab_code_formatter` tool can be used to format code cells to a consistent style: select each code cell, then click `Edit` and then one of the `Apply X Formatter` options (`YAPF` or `Black` are recommended).
- [NA] Include relevant tags in the final notebook cell (refer to the [DEA Tags Index](https://docs.dea.ga.gov.au/genindex.html), and re-use tags if possible)
- [NA] Clear all outputs, run notebook from start to finish, and save the notebook in the state where all cells have been sequentially evaluated
- [NA] Test notebook on both the `NCI` and `DEA Sandbox` (flag if not working as part of PR and ask for help to solve if needed)
- [NA] If applicable, update the `Notebook currently compatible with the NCI|DEA Sandbox environment only` line below the notebook title to reflect the environments the notebook is compatible with

### DIFF
--- a/Tools/dea_tools/bom.py
+++ b/Tools/dea_tools/bom.py
@@ -1,6 +1,23 @@
 """
-Scraping some BOM data
+Loading and processing Australian Bureau of Meteorology data.
+
+License: The code in this notebook is licensed under the Apache License,
+Version 2.0 (https://www.apache.org/licenses/LICENSE-2.0). Digital Earth
+Australia data is licensed under the Creative Commons by Attribution 4.0
+license (https://creativecommons.org/licenses/by/4.0/).
+
+Contact: If you need assistance, please post a question on the Open Data
+Cube Slack channel (http://slack.opendatacube.org/) or on the GIS Stack
+Exchange (https://gis.stackexchange.com/questions/ask?tags=open-data-cube)
+using the `open-data-cube` tag (you can view previously asked questions
+here: https://gis.stackexchange.com/questions/tagged/open-data-cube).
+
+If you would like to report an issue with this script, you can file one
+on Github (https://github.com/GeoscienceAustralia/dea-notebooks/issues/new).
+
+Last modified: March 2021
 """
+
 import datetime
 import pytz
 import ciso8601
@@ -114,10 +131,7 @@ def ui_select_station(stations,
                       center=(-24, 138),
                       **kw):
     """
-    Returns
-    =======
-
-    (gauge_data, station)
+    Create an interactive map for selecting river gauging stations.
     """
     import ipywidgets as W
     from IPython.display import display
--- a/Tools/dea_tools/landcover.py
+++ b/Tools/dea_tools/landcover.py
@@ -1,24 +1,22 @@
 # -*- coding: utf-8 -*-
 # Land_cover_plotting.py
 """
-Description: This file contains a set of python functions for plotting 
-Digital Earth Australia Land Cover data.
-License: The code in this notebook is licensed under the Apache License, 
-Version 2.0 (https://www.apache.org/licenses/LICENSE-2.0). Digital Earth 
-Australia data is licensed under the Creative Commons by Attribution 4.0 
+Plotting and animating Digital Earth Australia Land Cover data.
+
+License: The code in this notebook is licensed under the Apache License,
+Version 2.0 (https://www.apache.org/licenses/LICENSE-2.0). Digital Earth
+Australia data is licensed under the Creative Commons by Attribution 4.0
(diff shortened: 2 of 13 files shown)


Tools/dea_tools/bom.py
```diff
@@ -114,10 +131,7 @@ def ui_select_station(stations,
                       center=(-24, 138),
                       **kw):
     """
-    Returns
-    =======
-
-    (gauge_data, station)
+    Create an interactive map for selecting river gauging stations.
```
geoscience-aman, line 134: Suggestion: it might be good to specify the output and its data type.

Tools/dea_tools/landcover.py
```diff
@@ -273,12 +271,20 @@ def get_layer_name(measurement, da):
 
 def make_colorbar(fig, ax, measurement, horizontal=False, animation=False):
     """
-    Adds a new colorbar with appropriate land cover colours and labels
+    Adds a new colorbar with appropriate land cover colours and labels.
     
-    measurement = land cover measurement to use for colour map and lables
+    For level 4, must be used with a double plot, `ax` should be the left side canvas
+    the colour bar will go over the right hand side.
```
geoscience-aman: This sentence is hard to read for me and I'm not sure what exactly it means. Can I suggest splitting it up as something like the following:

"For level 4 data(?), this function must be used with a double plot. The 'ax' should be on the left side of the canvas (can we use figure instead of canvas?) and the colour bar will added on the right hand side."

Tools/dea_tools/spatial.py
```diff
@@ -1,7 +1,6 @@
 ## dea_spatialtools.py
 '''
-Description: This file contains a set of python functions for conducting 
-spatial analyses on Digital Earth Australia data.
+Tools for spatially manipulating Digital Earth Australia data (vectorising, rasterising, interpolating etc).
```
geoscience-aman: Perhaps "(vectorising, rasterising, interpolating etc)" is not needed to enhance conciseness. One can look at the list of functions available instead and infer that vectorisation, rasterisation and interpolation are available.

geoscience-aman (approved):
Thanks for your work on this @robbibt! I did not run or test the code with the changes. However, I have provided some comments after skimming through the file changes. The comments are mostly stylistic and a matter of preference. Hopefully the comments are helpful. I have also approved the pull request from my end :)

BexDunn (approved):
Have read through code and all seems good. Haven't tested the code to check that it works.

Tools/dea_tools/spatial.py
margaretharrison: done!

Tools/dea_tools/bom.py
robbibt: Hey @geoscience-aman, I 100% agree... the trouble is that with this function I'm not entirely sure what's being returned. I've gone down a rabithole and ended up here: 
https://github.com/Kirill888/jupyter-ui-poll/blob/develop/jupyter_ui_poll/_poll.py#L308-L309 
But that output doesn't really make sense to me. I think I might have to leave this for now and revisit later on (and probably improve documentation for a bunch of the functions in this module).

Tools/dea_tools/landcover.py
robbibt: Yep, that sounds much better. Updated!

robbibt:
Thanks so much for the reviews @margaretharrison + @geoscience-aman! Will merge now 😃

PR #968, opened 2022-11-02
Update of tide model paths to match new EFS location

robbibt:
### Proposed changes
Another update of the tide modelling file path in `model_tides` and `pixel_tides` to account for the new location of tide modelling files on the EFS path `/var/shared/tide_models`.

Have tested this on the dev Sandbox and the code works correctly.


vnewey (approved):
Looks great!

PR #969, opened 2022-11-02
Remove PyPI user from Github action

robbibt:
### Proposed changes
This attempts to fix the currently broken PyPI build action by removing `pypi_user` from the action.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -38,8 +38,8 @@ jobs:
         # uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
         uses: pypa/gh-action-pypi-publish@v1.4.2
         with:
-          # Password for your PyPI user or an access token
-          password: ${{ secrets.PYPI_PASS }}
+          # PyPI access token
+          password: ${{ secrets.PYPI_API_TOKEN }}
           # The target directory for distribution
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -38,8 +38,6 @@ jobs:
         # uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
         uses: pypa/gh-action-pypi-publish@v1.4.2
         with:
-          # PyPI user
-          user: ${{ secrets.PYPI_USER }}
           # Password for your PyPI user or an access token
           password: ${{ secrets.PYPI_PASS }}
```
pindge: this needs to be the token pass, is this  a token pass? better to get it right

as per doc, https://github.com/pypa/gh-action-pypi-publish, we should name the secret 

```
    password: ${{ secrets.PYPI_API_TOKEN }}
```

pindge (changes requested):

.github/workflows/publish.yml
Ariana-B: @pindge the naming is inconsistent across repos, in several places it's `secrets.PYPI_TOKEN`. I'm not sure if it really makes a difference at the end of the day, though deciding on a common convention would be nice

pindge (approved):

PR #973, opened 2022-11-09
Update coastal scripts and notebooks to use FES2014 tidal model

robbibt:
### Proposed changes
This PR updates DEA Notebooks to use the newly installed and set-up FES2014 tidal model. In the future, the plan is that this may replace the current tidal model we use in the Sandbox (OTPS), because it is difficult to install and has a dubious/unclear license. 

The PR updates tide modelling scripts inside `dea_tools.coastal` to use FES2014, but for continuity the older OTPS functions have been kept in the script but renamed with an "_otps" suffix (this will allow users to continue to use the older functions if they prefer).

The current `Tide_modelling.ipynb` notebook has been majorly refreshed to document these new functions and account for the change in tidal model, and the two Real world example notebooks (`Coastal_erosion.ipynb` and `Intertidal_elevation.ipynb`) updated to use the new funcs.

To test this PR, I think the easiest way might be to checkout this branch, then run the `Tide_modelling.ipynb` notebook from start to finish and see if anything breaks/doesn't make sense?


BexDunn (approved):
I can run the tidal modelling notebook, and it looks good to me.

margaretharrison (commented):
ignore this sorry!

erialC-P:
I worked through `Tide_modelling.ipynb` and it all looked good to me. The only thing that I noticed was a formatting issue: the cell outputs from the previous run were still present when I opened the notebook and need clearing. Otherwise, thanks for pushing this update through!

erialC-P (changes requested):
Content all looks great! Thanks for these updates Robbi! Only change required is a formatting one - the cell outputs could be cleared from the previous run.

robbibt:
> Content all looks great! Thanks for these updates Robbi! Only change required is a formatting one - the cell outputs could be cleared from the previous run.

Thanks Claire! I think our current policy is to run the entire notebook so that all the outputs are saved; that way they appear nicely formatted in the User Guide (there's always been an aspiration to find some way to wipe the outputs by the time they get loaded into the Sandbox, but I don't think we've every progressed that yet)

margaretharrison (approved):
Great work!! They all run for me!! I fixed up a few typos in the tidal modelling notebook - I hope I submitted them correctly. If not let me know and I'll try fix it!

robbibt:
Thanks so much @margaretharrison, good catches! If everyone is OK with it, I think I'll merge this PR now (happy to make any further changes as a new PR layer on too)

PR #972, opened 2022-11-08
Update legacy ARD to C3 ARD

geoscience-aman:
I updated the ARD product references from Legacy (`s2._ard_granule`) to Collection 3 (`ga_s2.m_ard_3`).

Due to other issues, I have excluded `'Real_world_examples/Sentinel_2_waterbodies.ipynb'` and `'Scientific_workflows/TSmask/TSmask.ipynb'` from this PR. I have tested all other affected notebooks on the Sandbox.

PR #974, opened 2022-11-10
Add C3 data to and remove C1 data from test datacube

geoscience-aman:
The test datacube needs to index C3 data to test the notebooks update to C3 data ([PR #972](https://github.com/GeoscienceAustralia/dea-notebooks/pull/972)). I have confirmed that the new indexed scenes exist.

### DIFF
--- a/Tests/setup_test_datacube.sh
+++ b/Tests/setup_test_datacube.sh
@@ -13,7 +13,7 @@ for metadata_yaml in $(find ./dea-config/product_metadata -name '*.yaml'); do
 done
 
 # Index products we care about for dea-notebooks
-for prod_def_yaml in $(find ./dea-config/products -name '*.yaml' -regex '.*\(ga_ls7e_nbart_gm_cyear_3\|ga_ls8c_nbart_gm_cyear_3\|ga_ls_wo_3\|s2a_ard_granule\|s2b_ard_granule\|ga_ls5t_ard_3\|ga_ls7e_ard_3\|ga_ls8c_ard_3\|ga_ls_wo_fq_cyear_3\).*'); do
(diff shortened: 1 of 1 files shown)


robbibt:
Hey @geoscience-aman , this looks fantastic! Can I make a suggestion though - can we do this update in the same PR as the other changes, e.g. #972? That way we can basically let the tests run and see if the updated notebooks pass right there (otherwise, these tests will fail because the current notebooks still use the older C1 products)

robbibt:
e.g. just copy out the changes to the `Tests/setup_test_datacube.sh` file from this PR and make them as a commit to the `c3_update` branch and see if the broken tests start running correctly again

geoscience-aman:
> Hey @geoscience-aman , this looks fantastic! Can I make a suggestion though - can we do this update in the same PR as the other changes, e.g. #972? That way we can basically let the tests run and see if the updated notebooks pass right there (otherwise, these tests will fail because the current notebooks still use the older C1 products)

Thanks @robbibt. My thinking was - why don't we just merge the tests first?

robbibt:
We totally can, it's just that it will be a bit tricky to work out if they're working as intended as they will fail by default without the updated notebooks... we'll then need to merge these changes into the `c3_update` anyway to see if it fixes the tests there, so just adding them there directly as a commit would cut out a few extra steps

geoscience-aman:
> We totally can, it's just that it will be a bit tricky to work out if they're working as intended as they will fail by default without the updated notebooks... we'll then need to merge these changes into the `c3_update` anyway to see if it fixes the tests there, so just adding them there directly as a commit would cut out a few extra steps

Oh gotcha. I missed that the tests need to be tested too haha.

robbibt:
Implemented in #972 instead, closing for now

PR #976, opened 2022-11-11
Replace provisional S3 C3 product to final

geoscience-aman:
Replace provisional S3 C3 products to final S3 C3 products. 

`ga_s2am_provisional_ard_3` -> `ga_s2am_ard_3`
`ga_s2bm_provisional_ard_3` -> `ga_s2bm_ard_3`

I have verified by eye that the outputs do not change for the changed notebooks. The automated 'GitHub Actions' check was  successful when it ran on the draft PR.


robbibt:
Awesome thanks @geoscience-aman ! I've asked a question about when the main S2 datasets get updated with NRT data on teams [here](https://teams.microsoft.com/l/message/19:8955cf633fb04fa58b1d67cc4cecef28@thread.tacv2/1668134005274?tenantId=645b613a-a64c-4250-975f-5f69787d028d&groupId=01e7a946-ba67-4594-bfb0-11aec6a7b50d&parentMessageId=1668134005274&teamName=DEA%20Projects&channelName=Sentinel%202%20Collection%203%20Upgrade&createdTime=1668134005274&allowXTenantAccess=false); depending on that I might jump into this branch and make some small changes to the `Exporting_satellite_images.ipynb` notebook.

robbibt (approved):
Hey @geoscience-aman , I've pushed a few changes to some of the scripts and notebooks here. Unfortuantely I think I misled you earlier - some of the notebooks aren't quite ready to drop provisional given that we don't currently have NRT data streaming into our main S2 C3 products. Because of this, I've had to restore provisional data in a few places:

- In the Exporting satellite imagery notebook; this can be removed as soon as we have NRT available in the main S2 C3 layers
- In the Sentinel-2 DEA datasets notebook, because it currently includes a big section on loading NRT data. This is temporary only; that whole notebook will likely need to be re-written from scratch once S2 C3 is ready.

The other change I made was to re-write the Burnt area mapping notebook so that it no longer needs to load NRT data at all - it just reads everything from the S2 C3 layer. This won't currently work if users want to use it to analyse latest satellite imagery (as there's no NRT available there yet), but should work seamlessly without any changes to the code as soon as the NRT pipeline starts working.

Thanks for all your work on this - I think we can merge this one for now and revisit those two notebooks later on. :)

Issue #656, opened 2020-10-26
Passing a dask array to dea_spatialtools.subpixel_contours raises an error


MatthewJA:
This:
```python
subpixel_contours(da.isel(time=0), z_values=[0.0])
```

Gives this:
```
~/dea-notebooks/Scripts/dea_spatialtools.py in subpixel_contours(da, z_values, crs, affine, attribute_df, output_path, min_vertices, dim, errors, verbose)
    453         contour_arrays = {str(i)[0:10]: 
    454                           contours_to_multiline(da, i, min_vertices)
--> 455                           for i in z_values}    
    456 
    457     else:

~/dea-notebooks/Scripts/dea_spatialtools.py in <dictcomp>(.0)
    453         contour_arrays = {str(i)[0:10]: 
    454                           contours_to_multiline(da, i, min_vertices)
--> 455                           for i in z_values}    
    456 
    457     else:

~/dea-notebooks/Scripts/dea_spatialtools.py in contours_to_multiline(da_i, z_value, min_vertices)
    413         # contour into a Shapely LineString feature
    414         line_features = [LineString(i[:,[1, 0]]) 
--> 415                          for i in find_contours(da_i.data, z_value)
    416                          if i.shape[0] > min_vertices]
    417 

/env/lib/python3.6/site-packages/skimage/measure/_find_contours.py in find_contours(array, level, fully_connected, positive_orientation, mask)
    134 
    135     segments = _get_contour_segments(array.astype(np.double), float(level),
--> 136                                      fully_connected == 'high', mask=mask)
    137     contours = _assemble_contours(segments)
    138     if positive_orientation == 'high':

skimage/measure/_find_contours_cy.pyx in skimage.measure._find_contours_cy._get_contour_segments()

/env/lib/python3.6/site-packages/skimage/measure/_find_contours_cy.cpython-36m-x86_64-linux-gnu.so in View.MemoryView.memoryview_cwrapper()

/env/lib/python3.6/site-packages/skimage/measure/_find_contours_cy.cpython-36m-x86_64-linux-gnu.so in View.MemoryView.memoryview.__cinit__()

TypeError: a bytes-like object is required, not 'Array'
```

But
```python
subpixel_contours(mndwi.isel(time=0).load(), z_values=[0.0])
```

works.

robbibt:
This has been fixed!

PR #978, opened 2022-11-14
Update and simplify `xr_rasterize`, `xr_vectorise` and `subpixel_contours` by using `odc-geo`

robbibt:
### Proposed changes
Several of our spatial functions (`xr_rasterize`, `xr_vectorise` and `subpixel_contours`) previously included complicated code like this for handling CRS and geotransform/affine information:

```
    # Check for a crs object
    try:
        crs = da.crs
    except:
        if crs is None:
            raise Exception("Please add a `crs` attribute to the "
                            "xarray.DataArray, or provide a CRS using the "
                            "function's `crs` parameter (e.g. 'EPSG:3577')")
            
    # Check if transform is provided as a xarray.DataArray method.
    # If not, require supplied Affine
    if transform is None:
        try:
            # First, try to take transform info from geobox
            transform = da.geobox.transform
        # If no geobox
        except:
            try:
                # Try getting transform from 'transform' attribute
                transform = da.transform
            except:
                # If neither of those options work, raise an exception telling the 
                # user to provide a transform
                raise TypeError("Please provide an Affine transform object using the "
                                "`transform` parameter (e.g. `from affine import "
                                "Affine; Affine(30.0, 0.0, 548040.0, 0.0, -30.0, "
                                "6886890.0)`")
```

This is clunky and prone to errors due to trying to handle all the different possible xarray datasets that users may pass in (e.g. data loaded from datacube, data loaded by rasterio etc).

Recently, the `odc-geo` package has provided some really nice ways to standardise access to CRS/transform info from any xarray. This lets us access this info consistently regardless of how an xarray dataset was originally loaded.

This PR updates the `xr_rasterize`, `xr_vectorise` and `subpixel_contours` functions to use `odc.geo` wrapped inside a simple helper func that ensures the data has a `.odc.geobox` method, greatly simplifying the CRS/geotransform handling code in these funcs:

https://github.com/GeoscienceAustralia/dea-notebooks/blob/88fe3599fd105950e2c2327aa8584af39d258fb5/Tools/dea_tools/spatial.py#L120-L121

This allows us to remove the `affine` param from these functions entirely, and should greatly reduce the likelihood of having to pass in a custom `crs` to these functions.

**To test:** Check out this branch, and try running `Frequently_used_code/Contour_extraction.ipynb` and `Frequently_used_code/Rasterize_vectorize.ipynb` from start to finish (or try passing in any weird/non-standard datasets to `xr_rasterize`/`xr_vectorise`/`subpixel_contours`)
 
### Closes issues (optional)
- Closes Issue #656 


cbur24 (approved):
Nice work @robbibt.  This is a large simplification of the existing `xr_rasterize` and `xr_vectorize` functions.  I ran the two notebooks in `Frequently_used_code` that reference these functions and they worked fine at my end.  I also rasterized and vectorized two dodgy datasets that didn't have good CRS info in their metadata and the functions worked as expected.

robbibt:
Thanks @cbur24, much appreciated!

PR #983, opened 2022-11-30
Add Sentinel-2 Collection 3 DEA datasets notebook

geoscience-aman:
Add a new Sentinel-2 notebook which contains new up-to-date information, including the background from the Sentinel-2 Collection 3 Content Management Interface (CMI) pages. 

To flag:

- Are we removing the old Sentinel-2 notebook (`DEA_datasets/Sentinel_2.ipynb`)? I have *not* included the removal of the old notebook as part of this PR.
- Sentinel-2B CMI links are not publicly accessible yet. Aaron followed up this morning as they are currently pending approval. 
- @robbibt will add `s2cloudless` as an option for Sentinel-2 in the `load_ard()` function. Update 16/01/2023: #984


uchchwhash (approved):
No objections from me!

robbibt (changes requested):
This is brilliant work @geoscience-aman, really nice clean code and great documentation! I've checked out the branch and made a few minor changes, main ones being:
- Adding a bit of extra info to the s2cloudless section, and changing the order the probability and masking layers are introduced
- Added a new section on filtering scenes according to their geometric accuracy

I really liked turning the Description section into a linked contents, so I did that for the Landsat C3 notebook too.

I still need to quickly test this on the NCI to see if it works there, but in the meantime can I get you to make two small changes? (you'll need to first pull the latest version of this branch to get my changes)

1. Yep, let's remove the existing `Sentinel2.ipynb` notebook from this folder - it will still be in Github's history if anyone needs it. If it's possible to do a quick search (not exactly sure how) of the repo to update any links that point to the old notebook, that would be fantastic.
2. I can't seem to find any documentation of this step for some reason, but to make the new notebook appear in the User Guide we need to add it to the index file here: https://github.com/GeoscienceAustralia/dea-notebooks/blob/create_s2_c3_notebook/DEA_datasets/README.rst . If you could put it up the top next to the Landsat one and remove the reference to the older S2 notebook that would be great!

robbibt (approved):

robbibt:
Looks great to me, happy to merge when you're ready! 😃

PR #987, opened 2023-01-09
Update interactive notebooks to switch to S2 C3 data instead of S2 Provisional

robbibt:
### Proposed changes
In a recent PR we excluded two interactive notebooks from being updated until NRT and final maturity data were contained in the same Sentinel-2 products:
https://github.com/GeoscienceAustralia/dea-notebooks/pull/976#pullrequestreview-1176832173

Because we now have a working dataset maturity workflow including both final and NRT, I have updated these remaining notebooks to use only the `ga_s2am_ard_3` and `ga_s2bm_ard_3` products.

(I also added Landsat 9 to both of these interactive notebooks to avoid us having to come back and add it later)


robbibt:
Hey @uchchwhash @geoscience-aman, this is a quick PR to remove Sentinel-2 Provisional data from the last two remaining interactive notebooks - I unfortunately formatted the data using Black for the first time so the diff is big, but I've tested both of the notebooks and they work nicely.

@JM-GA @ohehir FYI, apart from [Aman's S2 C3 notebook PR here](https://github.com/GeoscienceAustralia/dea-notebooks/pull/983) that's awaiting approval from Maree, this PR removes the last trace of S2 C1 data from the DEA Notebooks repo, so we can probably tick off that section of the decomissioning plan once these PRs are merged!

Tools/dea_tools/app/animations.py
```diff
@@ -162,7 +165,7 @@ def extract_data(self):
             "output_crs": crs,
             "group_by": "solar_day",
             "dask_chunks": {"time": 1, "x": 2048, "y": 2048},
-            "resampling": {"*": "cubic", "oa_fmask": "nearest", "fmask": "nearest"},
+            "resampling": {"*": "bilinear", "oa_fmask": "nearest", "fmask": "nearest"},
```
geoscience-aman, line 168: Hey @robbibt, just pointing this out to make sure the change from "cubic" to "bilinear" resampling was deliberate?

geoscience-aman (approved):
All looks good to me. Thanks @robbibt! 

I made just the one comment to confirm that the change from cubic to bilinear resampling in the `export_data()` function in `animations.py` was deliberate.

Tools/dea_tools/app/animations.py
robbibt: Good catch - yeah, "bilinear" is fine here (forgot I'd made that change), it's a bit quicker than "cubic" so no need to do that when the output is just for visualisation 😃

PR #988, opened 2023-01-09
Update misleading docstrings on coastal functions

robbibt:
### Proposed changes
Updates the newer coastal functions to correctly explain that they use the FES2014 tidal model, not OTPS/TPXO8.


erialC-P (approved):
Much clearer! :)

PR #1003, opened 2023-02-13
Add missing dependencies to setup.py, remove no longer used segmentation.py

robbibt:
### Proposed changes
In prep for a new version release of this package, I've cleaned up the `setup.py` file to add in some missing required packages to ensure it installs nicely on other systems (e.g. outside our NCI and Sandbox environment). 

As part of this, I removed the `segmentation.py` file from the repo as we no longer use it anywhere in DEA Notebooks since we stopped using `RSGISlib` for segmentation. It included a bunch of complex dependencies, so removing it will hopefully make installing `dea_tools` much easier for external users.


BexDunn (approved):
nice

PR #1008, opened 2023-03-01
simplify the test workflow

emmaai:
### Proposed changes
- Simplify the test workflow and make it reusable in `dea-sandbox`
- Remove `index` container, install the indexing tool in `sandbox` instead
- Make bash script pass the failure

### Closes issues (optional)
N/A

### DIFF
--- a/.github/workflows/test_notebooks.yml
+++ b/.github/workflows/test_notebooks.yml
@@ -32,6 +32,7 @@ jobs:
     - name: Set up Datacube and Test
       run: |
         sudo chown -R 1000:100 ./dea-notebooks
+        cd ./dea-notebooks 
         CURRENT_UID=1000:100 docker-compose up -d
         docker-compose exec -T sandbox ./dea-notebooks/Tests/setup_test_datacube.sh
         docker-compose exec -T sandbox ./dea-notebooks/Tests/test_notebooks.sh
--- a/.github/workflows/test_notebooks.yml
+++ b/.github/workflows/test_notebooks.yml
@@ -11,28 +11,13 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
-    - name: Set up containers
+    - name: Set up Datacube and Test
       run: |
         docker-compose up -d
-    - name: Set up Datacube
-      run: |
-        docker-compose exec -T index setup_test_datacube.sh
-        docker-compose exec -T index datacube product list
-    - name: Test with pytest
-      run: |
-        # First copy DEA Notebooks into the home directory where it lives in the Sandbox
-        docker-compose exec -T sandbox cp -r /dea-notebooks /home/jovyan/dea-notebooks
-
-        # Pip install `dea_tools` from the local Tools directory
-        docker-compose exec -T sandbox pip install -e /home/jovyan/dea-notebooks/Tools
-
-        # Run pytest using `--nbval-lax` which tests if notebookks run sucessfully or error 
-        docker-compose exec -T sandbox bash "-c" "cd /home/jovyan/dea-notebooks; pytest --durations=10 --nbval-lax Beginners_guide DEA_datasets Frequently_used_code/Contour_extraction.ipynb Frequently_used_code/Calculating_band_indices.ipynb Frequently_used_code/Downloading_data_with_STAC.ipynb Frequently_used_code/Exporting_GeoTIFFs.ipynb Frequently_used_code/Generating_composites.ipynb Frequently_used_code/Image_segmentation.ipynb Frequently_used_code/Opening_GeoTIFFs_NetCDFs.ipynb Frequently_used_code/Pansharpening.ipynb Frequently_used_code/Polygon_drill.ipynb Frequently_used_code/Principal_component_analysis.ipynb Frequently_used_code/Rasterize_vectorize.ipynb Frequently_used_code/Using_load_ard.ipynb Frequently_used_code/Virtual_products.ipynb"
-
-        # Shut down docker
-        docker-compose down
+        docker-compose exec -T sandbox ./dea-notebooks/Tests/setup_test_datacube.sh
+        docker-compose exec -T sandbox ./dea-notebooks/Tests/test_notebooks.sh
 
 #    - name: Setup upterm session
 #      uses: lhotari/action-upterm@v1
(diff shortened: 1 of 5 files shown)


robbibt:
Hey @emmaai! Is the test failure here expected? Or do you think it will start to work once it's merged?

emmaai:
no expected, should pass once able to pull the new sandbox image from ecr.

.github/workflows/test_notebooks.yml
```diff
@@ -7,32 +7,34 @@ on:
   pull_request:
     branches: [ develop, stable ]
 
+permissions:
+  id-token: write   # This is required for requesting the JWT
```
robbibt: What is JWT? Can we update the comment to clarify this?

docker-compose.yml
```diff
@@ -13,23 +13,8 @@ services:
...
-        - AWS_DEFAULT_REGION=ap-southeast-2
-     volumes:
-        - ./Tests/setup_test_datacube.sh:/usr/local/bin/setup_test_datacube.sh
-     depends_on:
-        - postgres
-     entrypoint: bash -c 'sleep infinity'
-
   sandbox:
-     image: geoscienceaustralia/sandbox:latest
+     image: 538673716275.dkr.ecr.ap-southeast-2.amazonaws.com/geoscienceaustralia/sandbox:latest
```
robbibt, line 17: I think this needs to point to the specific Sandbox image that is used on the "Default environment" on Prod Sandbox. E.g. `2.0.0`?

(We're going to have to update this every time we change the Prod Sandbox image, which isn't ideal - I'd love us to have a "stable" tag that we could use to set both the prod Sandbox image and this line too!)

Tests/setup_test_datacube.sh
```diff
@@ -1,12 +1,17 @@
 #!/usr/bin/env bash
 set -ex
+set -o pipefail
```
robbibt, line 5: What do these lines do? Can we add a code comment to make it clear for future readers?

.github/workflows/test_notebooks.yml
```diff
@@ -7,32 +7,34 @@ on:
...
+      uses: aws-actions/amazon-ecr-login@v1
 
-        # Shut down docker
-        docker-compose down
+    - name: Set up Datacube and Test
+      run: |
+        sudo chown -R 1000:100 ./dea-notebooks
+        CURRENT_UID=1000:100 docker-compose up -d
+        docker-compose exec -T sandbox ./dea-notebooks/Tests/setup_test_datacube.sh
+        docker-compose exec -T sandbox ./dea-notebooks/Tests/test_notebooks.sh
```
robbibt: Do we need to run `docker-compose down` here?

Tests/test_notebooks.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -ex
+set -o pipefail
```
robbibt, line 5: Same question as above

robbibt (changes requested):
Hi @emmaai, this looks really nice - I really like splitting out the tests into a sh script, and lots of great simplifications of the workflow!

I've added a few comments, mainly around some extra clarifications of some of the changes, and a likely change in Sandbox image to use.

Are the failed tests here OK? Do we need them to pass before we merge?

docker-compose.yml
emmaai: The whole point of setting up automated test workflow is to guarantee sandbox image not breaking notebooks in the update, or for us to be notified asap if it does require changes not back compatible. Any release will be based the current `latest` for docker image.  Hence if the test is performed on `latest`, it should be working on any image in the life cycle of `latest -> release -> latest`.

.github/workflows/test_notebooks.yml
emmaai: Everything will be killed when the workflow reaches the end.

.github/workflows/test_notebooks.yml
emmaai: JWT=json web token, as indicated by the name `id-token`. Thought it was a well-known term for AWS auth. (or was it not?)

emmaai: JWT=json web token, as indicated by the name `id-token`. Thought it was a well-known term for AWS auth. (or was it not?)

docker-compose.yml
robbibt: That's only one of the two use cases for our tests - they fullfill two different but important purposes on DEA Notebooks vs DEA Sandbox:

1) On DEA Sandbox: to guarantee changes to the Sandbox image don't break notebooks. For this, we need to test against `latest`.

2) On DEA Notebooks: to guarantee changes to Jupyter Notebooks and Python code don't break other code in the DEA Notebooks repo. For this, we need to test against **the same Sandbox image that is used for the "Default environment" on Prod Sandbox.** 

The #2 is really important for us - and testing against `latest` is useless for that because `latest` isn't available to general public users on the DEA Sandbox. We definitely need to run our tests against something like `2.0.0` for DEA Notebook's tests.

.github/workflows/test_notebooks.yml
robbibt: Thanks Emma - most DEA Notebooks maintainers (including myself) won't be familiar with AWS terminology. If we can expand this out to "JSON web token" in the code comment that would be great.

emmaai:
Ok, how about this, I’ll create a tag named `stable` for every release in
sandbox image, which will be the default on jupyterhub, for most of time
unless we need to roll back when broken. Then we don’t need to change the
tag in the workflow yaml any more.

robbibt:
> @emmaai Ok, how about this, I’ll create a tag named `stable` for every release in sandbox image, which will be the default on jupyterhub, for most of time unless we need to roll back when broken. Then we don’t need to change the tag in the workflow yaml any more.

I think that would be fantastic!

.github/workflows/test_notebooks.yml
robbibt:
```suggestion
  id-token: write   # This is required for requesting the JSON web token
```

docker-compose.yml
emmaai: Can we merge as it is though, for the time being? I have some changes to make to `dea-sandbox` after merging this pr. I'll add new tag together.  Then I can come back again and change here when the image is ready to pull.

robbibt: Yep sounds good, will mark as approved now

robbibt (approved):
Looks great, happy to merge and then update the image tag later. Thanks Emma!

PR #1028, opened 2023-04-06
Fix `pixel_tides` issues when `ds` has no time dimension

robbibt:
### Proposed changes
This PR updates the `pixel_tides` function to better handle situations where an input array has no time dimension. 

The function failed in this scenario:

- You pass in a dataset `ds` with no time dimension, and a set of custom times via the `times` param
- The function models tides into a pandas dataframe, then tries to reshape and transpose the data back into 3D xarray format by matching it to the original dimensions from `ds`
- Because the modelled tides dataframe has multiple timesteps but the original `ds` has no time dim, the [transpose steps fails](https://github.com/GeoscienceAustralia/dea-notebooks/compare/pixel_tides_update?expand=1#diff-9a8d3841a66e6cc352119d7194f661f0ac4be96b059c0ea527640d30d65e55d9L605) when trying to reshape x/y/time data into x/y dimensions

image

Two fixes to address this:

- Better error catching to make the function fail completely if a dataset with no time dimension is passed and no custom times are provided
- Hard code the list of dimensions instead of getting them from the input `ds`. This ensures they are in the order expected by the re-projection step.

image

Have also cleaned up some un-needed OTPS code in one of the new FES2014 funcs.

### DIFF
--- a/Tools/dea_tools/coastal.py
+++ b/Tools/dea_tools/coastal.py
@@ -16,7 +16,7 @@ https://gis.stackexchange.com/questions/tagged/open-data-cube).
 If you would like to report an issue with this script, you can file one 
 on Github (https://github.com/GeoscienceAustralia/dea-notebooks/issues/new).
 
-Last modified: March 2023
+Last modified: January 2023
 
 """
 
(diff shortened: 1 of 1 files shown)


erialC-P (approved):
Good catch Robbi! This all looks thorough and ready to merge :)

PR #1030, opened 2023-04-11
Fix geocoding func to still work if geocoding service is down

robbibt:
### Proposed changes
A small fix while we're waiting to merge `develop` into `stable` for the directory re-structure: I noticed that the interactive app notebooks are currently failing because the geocoding service is down. This PR allows the function to work even when this happens, by returning coordinates instead of a geocoded place name.


erialC-P (approved):
Good catch and neat solution! Approved :)

PR #1042, opened 2023-04-20
Additional error handling for plot_variable_images tool in plotting.py

Matt-dea:
Added extra check to handle error when the input xarray dataset has 0 images.

### Proposed changes
Added additional code block to handle errors when the input dataset has 0 images.

### DIFF
--- a/Tools/dea_tools/plotting.py
+++ b/Tools/dea_tools/plotting.py
@@ -1078,16 +1078,16 @@ def plot_variable_images(img_collection):
 
     # Calculate number of images in `img_collection`
     plot_count = img_collection.dims["time"]
-        
+    
     # Check if dataset has 0 images
     if plot_count == 0:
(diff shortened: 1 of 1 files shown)


supermarkion (approved):
We add a new if-else statement to handle the corner case. Looks good to me.

robbibt:
Hey @Matt-dea, can we raise an error here rather than return a string? Something like e.g.


```
  # Check if dataset has 0 images
  if plot_count == 0:
      if hasattr(img_collection, "sensor"):
          raise ValueError("The {} dataset has no images to display for the " 
          "given query parameters".format(img_collection.sensor))
      else:
          raise ValueError("The supplied xarray dataset has no images to "
          "display for the given query parameters")
```

Matt-dea:
Great suggestion Robbie! I agree. Changes implemented and passed checks. I will merge to develop.

PR #1043, opened 2023-04-21
Add Sentinel-2 and Landsat option to animation notebook

robbibt:
### Proposed changes
A small PR to add a combined "Sentinel-2 and Landsat" layer to our interactive animations notebook. This will allow users to create animations that combine both Sentinel-2 and Landsat data:

image

### DIFF
--- a/Interactive_apps/Generating_satellite_animations.ipynb
+++ b/Interactive_apps/Generating_satellite_animations.ipynb
@@ -79,6 +79,7 @@
    ]
   },
   {
+   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


geoscience-aman (approved):

Matt-dea:
<p>Hi Robbi, awesome update. I can't find any changes to suggest.</p><p>I ran the notebook in the sandbox and the mixed product worked fine. I did receive an error that I did not get when running a single product animation, but it didn't stop the animation from being produced. Pretty sure nothing to worry about....</p><pre class="ql-syntax" spellcheck="false">"/env/lib/python3.8/site-packages/rasterio/warp.py:344: NotGeoreferencedWarning: Dataset has no geotransform, gcps, or rpcs. The identity matrix will be returned.
  _reproject(
</pre><p> </p><p> </p>

robbibt:
Yeah, I don't know what that one means - it's been showing up a lot recently but I haven't worked out why/when it occurs. Doesn't seem to break anything, but it's annoying!

PR #1055, opened 2023-05-16
Stable update for DEA Notebooks showcase

robbibt:
### Proposed changes
This PR merges recent changes to the `stable` branch of DEA Notebooks so that they will become available in the DEA Sandbox and be published on the DEA Docs.


JM-GA (approved):
Thanks @robbibt!

PR #1062, opened 2023-05-18
Add temporary fix for ipywidget > 8.0 fileupload changes

robbibt:
### Proposed changes
A stakeholder has got in touch to let us know the file upload functionality on the Coastal Transects app is broken, failing with a "No transect drawn or uploaded. Please select a transect on the map, or upload a GeoJSON or ESRI Shapefile" error when you try to upload a GeoJSON file or ESRI Shapefile.

This appears to be an issue with a change to `ipywidgets` version 8 in the latest Sandbox image, which has an API change here:
https://ipywidgets.readthedocs.io/en/latest/user_migration_guides.html#fileupload

This PR implements a quick fix using a suggested migration from that doc. We can switch to the new `fileupload` widget API later.

### DIFF
--- a/Tools/dea_tools/app/deacoastlines.py
+++ b/Tools/dea_tools/app/deacoastlines.py
@@ -284,11 +284,16 @@ class transect_app(HBox):
 
         # Clear any drawn data if present
         self.gdf_drawn = None
+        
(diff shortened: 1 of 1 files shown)


vnewey (approved):
Your fix works a treat @robbibt!

PR #1065, opened 2023-05-22
Allow `subpixel_contours` time format to be customised

robbibt:
### Proposed changes
A small PR, mainly to test whether our new tests allow us to successfully add features to `dea-tools` functions without breaking existing code. The new tests successfully caught several issues with the PR, and pass now that they were fixed!

The specific change allows users to customise how dates are formatted in a geopandas object returned by the `subpixel_contours` function, e.g.:

image

For example, a user may want dates to be formatted like `"%Y-%m-%d"` (the default), or like `"%Y"`, `"%Y-%m"` etc. This change replaced the previous approach of crudely converting dates to a string then clipping them to the first 10 characters (`str(i)[0:10]`). 

(This update will be required for any future updates to DEA Coastlines that require sub-annual shorelines)

### DIFF
--- a/Tools/dea_tools/spatial.py
+++ b/Tools/dea_tools/spatial.py
@@ -418,7 +418,7 @@ def subpixel_contours(
         """
         if isinstance(i, np.datetime64):
             ts = pd.to_datetime(str(i))
-            i = ts.strftime(date_format)
+            i = ts.strftime(time_format)
         return i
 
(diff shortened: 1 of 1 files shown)


erialC-P (approved):
Nice one Robbi - looks good to me! Always good to have options with dates and times. It's such a curse mucking around with date formats.

PR #1067, opened 2023-05-30
fix remote zipped shapefile reading

emmaai:
### Proposed changes
Download remote zipped shapefile before reading

Reasons
- OGR is unable to read the remote zipped shapefile when .shp is under a subfolder
- server doesn't support range (chunk) reading of a binary file
### Closes issues (optional)
N/A


robbibt:
Hey @emmaai, thanks so much for this. This issue has happened before whenever the ACT gov makes changes to that file. Because of that, I think in this case we might implement a simpler solution by simply uploading the vector file into the `Supplementary_data` folder (it's [CC BY 4.0](https://data.gov.au/data/dataset/act-suburb-locality-boundaries-geoscape-administrative-boundaries) so we can just re-host it). That way the file is always there, and we can avoid anything like this happening in the future.

I'll push an update to this branch and merge when it passes tests!

robbibt (approved):

PR #1074, opened 2023-06-14
Update the script that generates training data to work with field poi…

tebadi:
Update the existing script for generating training data to allow for matching field data with ARD data that fall within a certain time range.

### DIFF
--- a/Tools/dea_tools/classification.py
+++ b/Tools/dea_tools/classification.py
@@ -52,7 +52,7 @@ from sklearn.model_selection import KFold, ShuffleSplit
 from sklearn.model_selection import BaseCrossValidator
 
 from datacube.utils.geometry import assign_crs
-from datacube.utils import geometryblack
+from datacube.utils import geometry
 from dea_tools.spatial import xr_rasterize
 
 
--- a/Tools/dea_tools/classification.py
+++ b/Tools/dea_tools/classification.py
@@ -27,10 +27,12 @@ import numpy as np
 import pandas as pd
 import xarray as xr
 import time
+import warnings
 from datetime import timedelta
 import geopandas as gpd
 from copy import deepcopy
 from tqdm.auto import tqdm
+import multiprocessing as mp
 import matplotlib.pyplot as plt
 
 from typing import Callable, Tuple, Any, Optional, List, Dict
@@ -40,24 +42,17 @@ from dask_ml.wrappers import ParallelPostFit
 import dask.distributed as dd
 from dask.diagnostics import ProgressBar
 
-import multiprocessing as mp
-
 from sklearn.cluster import KMeans
 from sklearn.utils import check_random_state
 from abc import ABCMeta, abstractmethod
-from datacube.utils import geometry
 from sklearn.base import ClusterMixin
-
-from rasterio.features import rasterize
-
 from sklearn.mixture import GaussianMixture
-from datacube.utils.geometry import assign_crs
 from sklearn.cluster import AgglomerativeClustering
 from sklearn.model_selection import KFold, ShuffleSplit
 from sklearn.model_selection import BaseCrossValidator
 
-import warnings
-
+from datacube.utils.geometry import assign_crs
+from datacube.utils import geometryblack
 from dea_tools.spatial import xr_rasterize
 
 
@@ -190,7 +185,10 @@ def sklearn_unflatten(output_np, input_xr):
     output_xr = xr.DataArray(
         output_ma,
         coords={"z": stacked["z"]},
-        dims=["z", *["output_dim_" + str(idx) for idx in range(len(output_px_shape))]],
+        dims=[
+            "z",
+            *["output_dim_" + str(idx) for idx in range(len(output_px_shape))],
+        ],
     )
 
     output_xr = output_xr.unstack()
@@ -280,7 +278,9 @@ def predict_xr(
 
     # set chunk size if not supplied
     if chunk_size is None:
-        chunk_size = int(input_xr.chunks["x"][0]) * int(input_xr.chunks["y"][0])
+        chunk_size = int(input_xr.chunks["x"][0]) * int(
+            input_xr.chunks["y"][0]
+        )
 
     def _predict_func(model, input_xr, persist, proba, clean, return_input):
         x, y, crs = input_xr.x, input_xr.y, input_xr.geobox.crs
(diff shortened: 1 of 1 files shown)


Tools/dea_tools/classification.py
```diff
@@ -23,35 +23,36 @@
...
 from sklearn.mixture import GaussianMixture
-from datacube.utils.geometry import assign_crs
 from sklearn.cluster import AgglomerativeClustering
 from sklearn.model_selection import KFold, ShuffleSplit
 from sklearn.model_selection import BaseCrossValidator
 
-import warnings
-
+from datacube.utils.geometry import assign_crs
+from datacube.utils import geometryblack
```
robbibt: Is this meant to be just "geometry"?

Tools/dea_tools/classification.py
```diff
@@ -944,15 +1026,19 @@ def spatial_clusters(
         Index of the cluster each sample belongs to.
     """
     if method not in ["Hierarchical", "KMeans", "GMM"]:
-        raise ValueError("method must be one of: 'Hierarchical','KMeans' or 'GMM'")
+        raise ValueError(
+            "method must be one of: 'Hierarchical','KMeans' or 'GMM'"
```
robbibt:
```suggestion
            "Method must be one of: 'Hierarchical','KMeans' or 'GMM'"
```

Tools/dea_tools/classification.py
```diff
@@ -445,25 +480,39 @@ def _get_training_data_for_shp(
...
     data = feature_func(dc_query)
 
-    # create polygon mask
-    mask = xr_rasterize(gdf.iloc[[index]], data)
-    data = data.where(mask)
+    # if no data is present then return
+    if len(data) == 0:
+        return
+
+    if gdf.iloc[[index]].geometry.geom_type.values != "Point":
```
robbibt, line 506: Awesome!

Tools/dea_tools/classification.py
```diff
@@ -637,11 +697,17 @@ def feature_function(query):
     max_retries: int, default 3
         Maximum number of times to retry collecting samples. This number is invoked
         if the 'fail_threshold' is not reached.
+    time_field: str
```
robbibt, line 700: This sounds really useful!

For our Coastal Ecosystems project, we're using the same time range for all our training data (`time="2021"`). Can I confirm that if we leave `time_field` and `time_delta` to `None`, we can still provide our constant time period in our `dc_query` dict?

Tools/dea_tools/classification.py
```diff
@@ -23,35 +23,36 @@
...
-from dask_ml.wrappers import ParallelPostFit
 from sklearn.mixture import GaussianMixture
-from datacube.utils.geometry import assign_crs
 from sklearn.cluster import AgglomerativeClustering
 from sklearn.model_selection import KFold, ShuffleSplit
 from sklearn.model_selection import BaseCrossValidator
 
-import warnings
-
+from datacube.utils.geometry import assign_crs
```
robbibt, line 54: I'd love to switch all these funcs to use `odc.geo` instead of importing from `datacube.utils.geometry`, but I think that might be a job for a new PR

robbibt (changes requested):
Hey @tebadi, this looks awesome! I've left a few small comments below. A few questions:
1) Have you tried running the notebooks located here (https://docs.dea.ga.gov.au/notebooks/Real_world_examples/Scalable_machine_learning/README.html) to make sure they still work after these updates? They aren't yet included in our automated tests, so need to be manually run at this stage (if they don't work, we can easily update the notebooks themselves)
2) I'd really love to get some of these funcs under test coverage. Thoughts on adding a `test_classification.py` test script here (https://github.com/GeoscienceAustralia/dea-notebooks/tree/develop/Tests/dea_tools) with some basic unit tests, perhaps run off a really tiny sample training data file? Could also save this for a future PR, but it could be good to start thinking about how we can make that happen

Tools/dea_tools/classification.py
tebadi: Yes :) This is fixed now.

Tools/dea_tools/classification.py
tebadi: Yes. These are the lines that update the `time` element of an odc query:

https://github.com/GeoscienceAustralia/dea-notebooks/blob/591feb9f30214b8f62466d632ef699e30575b004/Tools/dea_tools/classification.py#L491

If time_delta isn't provided, then the query timestamp doesn't get updated and remains as the constant value that you provided in the query.

Tools/dea_tools/classification.py
tebadi: This will be a separate PR.  I'm happy to do this so feel free to create a ticket and assign it to me.

tebadi:
> Hey @tebadi, this looks awesome! I've left a few small comments below. A few questions:
> 
> 1. Have you tried running the notebooks located here (https://docs.dea.ga.gov.au/notebooks/Real_world_examples/Scalable_machine_learning/README.html) to make sure they still work after these updates? They aren't yet included in our automated tests, so need to be manually run at this stage (if they don't work, we can easily update the notebooks themselves)
> 2. I'd really love to get some of these funcs under test coverage. Thoughts on adding a `test_classification.py` test script here (https://github.com/GeoscienceAustralia/dea-notebooks/tree/develop/Tests/dea_tools) with some basic unit tests, perhaps run off a really tiny sample training data file? Could also save this for a future PR, but it could be good to start thinking about how we can make that happen

1- I know some of them fail after the datacube upgrade (not related to my changes).  The base extract script fails with the same error that we reported in odc-geo.  I'll have to add the crs somehow.  I will need to do this since I will need the script for extracting data for the cultivated model (geometry shape is polygon from memory but it still failed).  This is in my todo list.

2- I'd love to add some unit tests here now I'm using this widely for different products.  Feel free to add a ticket and assign it to me otherwise I'll ask Cedric to add this to our board.

robbibt (approved):
@tebadi I ran the Scalable Machine Learning notebooks and they work fine for me on the new code, so I think this is ready to go!

I'll create some extra issues for the things discussed in this thread and assign them to you.

Thanks for the updates to this - feel free to squash and merge when ready!

tebadi:
> xtra issues for the things discussed in this thread and assign them to y



> @tebadi I ran the Scalable Machine Learning notebooks and they work fine for me on the new code, so I think this is ready to go!
> 
> I'll create some extra issues for the things discussed in this thread and assign them to you.
> 
> Thanks for the updates to this - feel free to squash and merge when ready!

I might have some local changes - thanks for checking this!

Issue #989, opened 2023-01-10
Update tide modelling functions to use refactored version 2.0 of `pyTMD`


robbibt:
https://github.com/tsutterley/pyTMD/releases/tag/2.0.0

2320sharon:
Hi @robbibt 
Thank you for all your hard work on the dea_notebooks and dea_tools package. I see you're planning on updating to pyTMD version 2. We are working on a package that might use the dea_tools package to tidally correct satellite imagery and I was wondering what your estimated time frame was on updating to version 2 of pyTMD?

robbibt:
Hey @2320sharon! I'll look into it this week and get a better idea of how complex this will be - I don't anticipate it will be too tricky. We're actually hoping to release the first official version 1.0.0 of `dea_tools` over the next week or so, so I'll see if we can get the pyTMD update into that release. 

Either way though, we'll aim to make the update in a backwards-compatible way, so I don't think the switch to pyTMD 2.0 should change much about how the tide modelling funcs in `dea_tools` behave, so you can probably use one of the [latest versions here](https://pypi.org/project/dea-tools/#history) for now without much issue. The pyTMD 2.0 updates may make things a bit more efficient to run, but so far the biggest impact on performance we've found is to clip the FES2014 tidal consituent NetCDFs to a bounding box around your study area (e.g. we clip them to Australia) rather than use the global NetCDFs out of the box (we've found it can reduce the time to load those consituents in pyTMD from minutes to just a few seconds). 😃

2320sharon:
Thank you for taking the time to dig into this issue this week and for responding so quickly. It's great news to hear that the first official release of `dea_tools` is almost ready to go, and it's perfectly timed!💯 

 I also appreciate your clarification that the new version of dea_tools will likely be backwards compatible. Initially, I was concerned that the significant differences in the code structure between `pyTMD 1.1.3` and `pyTMD 2.0.0` would require substantial modifications to `dea_tools`. I'm planning to investigate  the [latest versions here](https://pypi.org/project/dea-tools/#history) of the code and see it will work for our use case.

robbibt:
[...]

PR #1077, opened 2023-06-16
Update tide modelling functions to use `pyTMD` version 2.0

robbibt:
### Proposed changes
The Python package we use for tide modelling (`pyTMD`) recently went through a major upgrade to version 2. This PR updates our coastal tide modelling tools to make them work on the latest version of `pyTMD`.

The changes pass our DEA Notebooks tests defined here, including several tests that make sure the updated functions are still as accurate as before: https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Tests/dea_tools/test_coastal.py

From the passed tests, it looks like upgrading to `pyTMD` version 2 hasn't affected any of our functions or reduced any of our tide modelling accuracy. 😃 

### Closes issues (optional)
- Closes Issue #989 


erialC-P (approved):
Looks great to me! Thanks for adding the new version :)

PR #1080, opened 2023-06-19
Minor updates to new turbidity notebook

robbibt:
### Proposed changes
This PR makes a few additional small updates to the turbidity notebook:
- Simplified some of the plotting code (using xarray's `.plot.imshow` directly instead of `plt.imshow(...)`
- Reduced the size of the figures and animations to make the notebook smaller (now 24 mb vs 65 mb)
- Since this is primarily a visualisation notebook, reduced the analysis resolution down to 60 x 60 m to reduce memory usage and processing time. Hopefully this is OK - users can easily increase it themselves if they want higher-res updates, but the outputs still look really nice!

https://github.com/GeoscienceAustralia/dea-notebooks/assets/17680388/9b3d0813-2f85-4ed1-b425-9ce2be7185c9

- Have loaded the data into memory earlier on (just after the data load). Normally it can be good to keep things out of memory as long as possible, but here there's so many plotting cells that the same data was being loaded into memory again and again, making things slower than necessary. Now this only happens once, up the very top.
- Added the missing "Additional information" section at the bottom
- Added notebook to the README so it will render in the docs

### DIFF
--- a/Real_world_examples/README.rst
+++ b/Real_world_examples/README.rst
@@ -19,6 +19,7 @@ More complex case study-based workflows demonstrating how DEA can be used to add
    Seasonal_water_extents.ipynb
    Shipping_lane_identification.ipynb
    Surface_area_duration.ipynb
+   Turbidity_animated_timeseries.ipynb
    Urban_change_detection.ipynb
    Vegetation_phenology.ipynb
    Wetness_stream_gauge_correlations.ipynb
--- a/Tools/dea_tools/datahandling.py
+++ b/Tools/dea_tools/datahandling.py
@@ -17,7 +17,7 @@ here: https://gis.stackexchange.com/questions/tagged/open-data-cube).
 If you would like to report an issue with this script, you can file one
 on Github (https://github.com/GeoscienceAustralia/dea-notebooks/issues/new).
 
-Last modified: February 2023
+Last modified: June 2023
 """
 
 # Import required packages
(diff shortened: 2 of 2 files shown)


robbibt:
And @Joram95 - have included an author credit here, hope this is OK!
image

erialC-P (commented):
Thanks Robbi. I forgot to add the bottom section `Additional info` - I thought something was missing :) The resampling is a good idea too.
This looks great to me. One question though, in the final animation, a colour bar appears at the bottom with a narrow range of NDTI values on it and these dont change as the animation progresses. Is this expected? Do we also need to set a consistent range on the the colour ramp across the whole image set to something like -0.5 to 0.5?

Joram95:
Thanks @robbibt and @erialC-P,
Unfortunately reviewing the code on a phone is not possible. I'm confident to go with all your changes.

I also have those same thoughts about the scale bar @erialC-P.

Regarding authorship please put your own names down as authors, I did use a lot of your code from the 'real world examples' and you both have been great helps.

robbibt:
@erialC-P @Joram95 Easily done - set the scale bar to a consistent -0.5 to 0.0 (not 0.5, as most of the interesting stuff was happening below 0.0). Hope this looks OK!

https://github.com/GeoscienceAustralia/dea-notebooks/assets/17680388/2074de39-51e3-42e7-971e-66971ecb28aa

And thanks Joram! We'll chat about our approach to notebook authors internally (particularly around DEA staff) and work out the best way to specify all contributors - for now you definitely did the vast majority of the work so I think it's definitely worth having your name attached to the notebook!

robbibt:
And @erialC-P, I popped in the original version of `datahandling` and updated the docstrings. Should be ready for review again!

erialC-P (approved):
Love it! Thanks heaps Robbi

Issue #975, opened 2022-11-10
problem with plot land cover tool


Kooie-cate:
**Describe the bug/issue**
A clear and concise description of what the bug/issue is, including screenshots if relevant. 

**Steps to reproduce**
Steps to reproduce the behaviour. For example:
1. Open DEA Sandbox
2. Launch the "DEA_datasetsDEA_land_cover.ipynb" notebook
3. change time range in cell 4 to be a single year
4. Run all cells except skip cell 10
5. Code in section "11" produces error "AttributeError: 'AxesSubplot' object has no attribute 'flat'"

**Expected behaviour**
I would expect a single temporal layer to simply plot the same as it does when the "year" variable is given. the code is trying to generate a plot with multiple subplots then breaking when trying to fit a colour bar to one of those subplots.

**Environment information**
[DEA Sandbox]

robbibt:
Closed by #1095

PR #1095, opened 2023-06-30
Lc plot fix

Kooie-cate:
### Proposed changes
This pull request includes changes to dea_tools/landcover/plot_land_cover()
fixes error described in issue #975, where dataset with a single time step would not plot neatly. new protocols introduced to handle this correctly.

also included is a change to DEA_products/DEA_land_cover.ipynb
this change removes code that no-longer works with maplotlib update to customise colour bar tick marks, when called from xr.dataarray.plot(). 

No new notebooks added, 

the use of plot_land_cover() has been checked in the three notebooks where it is called (DEA_land_cover.ipynb, How_to_guides/Land_cover_animated_plots.ipynb and How_to_guides/Land_cover_change_mapping.ipynb) and they are not effected.

### Closes issues (optional)
- Closes Issue #975

- [X ] Notebook created using the [DEA-notebooks template](https://github.com/GeoscienceAustralia/dea-notebooks/tree/develop)
- [X ] Remove any unused Python packages from `Load packages`
- [ X] Remove any unused/empty code cells
- [ X] Remove any guidance cells (e.g. `General advice`)
- [ X] Ensure that all code cells follow the [PEP8 standard](https://www.python.org/dev/peps/pep-0008/) for code. The `jupyterlab_code_formatter` tool can be used to format code cells to a consistent style: select each code cell, then click `Edit` and then one of the `Apply X Formatter` options (`YAPF` or `Black` are recommended).
- [ X] Include relevant tags in the final notebook cell (refer to the [DEA Tags Index](https://docs.dea.ga.gov.au/genindex.html), and re-use tags if possible)
- [X ] Clear all outputs, run notebook from start to finish, and save the notebook in the state where all cells have been sequentially evaluated
- [ X] Test notebook on both the `NCI` and `DEA Sandbox` (flag if not working as part of PR and ask for help to solve if needed)
- [ X] If applicable, update the `Notebook currently compatible with the NCI|DEA Sandbox environment only` line below the notebook title to reflect the environments the notebook is compatible with


DEA_products/DEA_Land_Cover.ipynb
erialC-P, line 1: in `plot_layer(LEVEL3_COLOUR_SCHEME, lc.level3)`, the old version used to print out classification labels in the legend but the new version has replaced these with the numeric classes - is this expected behaviour?

erialC-P (approved):
Looks great Cate! I've left a single comment only, questioning whether class labels are meant to print out in the first call to `plot_layer` but otherwise this looks good to go to me :)

DEA_products/DEA_Land_Cover.ipynb
Kooie-cate: yes, i know it's not ideal but the functionality that enabled me to add string labels the the colour bar of the plot is not there anymore.

PR #1096, opened 2023-06-30
Add search bar and marker to interactive widgets.

geoscience-aman:
### Proposed changes
The addition of a search bar means that users don't have to find a location by dragging and zooming manually. 

Added with the exact same arguments to both `Tools/dea_tools/app/imageexport.py` and `Tools/dea_tools/app/animations.py` for consistency.

Screenshot of it in action:

image


Kooie-cate (approved):
Hi Aman, this works really well in both apps, top job

PR #1101, opened 2023-07-20
fix rename param in xr rasterize

FlorisCalkoen:
### Proposed changes
The rename parameter of the xr_rasterize function was not working because the renamed variable was not saved.

This
```python
da_rasterized = odc.geo.xr.wrap_xr(im=im, 
                                       gbox=da.odc.geobox)
    da_rasterized.rename(name)
```

Is changed to: `da_rasterized = odc.geo.xr.wrap_xr(im=im, gbox=da.odc.geobox).rename(name)`

In a new PR to avoid conflicts due to black formatting.

### DIFF
--- a/Tools/dea_tools/spatial.py
+++ b/Tools/dea_tools/spatial.py
@@ -2,18 +2,18 @@
 '''
 Tools for spatially manipulating Digital Earth Australia data.
 
-License: The code in this notebook is licensed under the Apache License, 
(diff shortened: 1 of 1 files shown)


BexDunn:
Hey @FlorisCalkoen , thanks for this fix! We have added it into the latest version of our code. Can you please try pulling the latest version of the develop branch into your fork? I think we have applied this fix here, please let us know if you have questions:  https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Tools/dea_tools/spatial.py#L284

PR #1112, opened 2023-08-28
Major refactor of tide modelling funcs to add improved parallelization

### LINKED ISSUE #1109: Provide a global method to set the tidal data directory
The `dea_tools.coastal.model_tides()` function has a `directory` argument which defaults to `/var/share/tide_models`. This works fine when running or testing `dea_tools` in DEA Sandbox or inside a Docker container, but is awkward anywhere else.

I don't want to overcomplicate things, but would like to do something like allowing changing this default with an Environment variable, to allow a bit more flexibility.

For example, I could create a `~/tide_models` directory in my home directory, and set `DEA_TOOLS_TIDE_MODELS` to point to it.

robbibt:
### Proposed changes
This PR implements a major refactor of the `model_tides` and `pixel_tides` functions to improve performance when modelling large numbers of points/timesteps with multiple models by adding parallelisation using `concurrent.futures` and Dask. 

Parallelisation in `model_tides` is applied both across each individual tide model, as well as (optionally) by splitting tide modelling points into smaller chunks that are run in parallel. These are configurable via the `parallel` and `parallel_splits` params. 

For modelling **10,000 timesteps, 1000 lat/lon points and 6 tide models** on the 16XL DEA Sandbox server with `parallel_splits=10`, this improves runtimes by more than an order of magnitude:

- **Before:** 29 min, 22 sec
- **After:** 1 min, 16 sec

The `pixel_tides` function has additionally been updated to use new Dask reproject functionality from `odc-geo`, used to reproject low-res tide outputs to each higher res satellite pixel (@Kirill888).

Other updates:
- Allow the tide modelling file directory to be optionally defined via a global environmental variable `DEA_TOOLS_TIDE_MODELS` (thanks to a suggestion by @omad in #1109)
- Allow results from `model_tides` to be returned in wide or long dataframe format
- Allow the user to control the units of modelled tides (e.g. to improve memory by outputting results as integer centimetres vs floating point metres)
- Expand `coastal.py` test coverage to include newly added multiple tide modelling functionality
- Added some extra words to the spellcheck dictionary

### Closes issues
Closes #1109


robbibt:
Hey @2320sharon @dbuscombe-usgs, this might be of interested! 🌊

erialC-P (approved):
Hey Robbi! Everything looks hunky-dory to me :) Nice work on the optimisation!!!

2320sharon:
This looks awesome! I'm definitely going to try these optimizations out. Thanks for this!
> Hey @2320sharon @dbuscombe-usgs, this might be of interested! 🌊

PR #1123, opened 2023-09-05
Make dask configurable inside `pixel_tides`

robbibt:
### Proposed changes
Currently the `pixel_tides` function uses Dask under the hood to resample our low resolution tide modelling outputs into the original higher-res pixels of our satellite data. Previously, we used `.compute()` inside the function to load our results into memory and return then as a normal `xr.DataArray`.

However, for certain applications it can be useful to return our tide modelling data as an un-computed Dask array, for example to use as part of a larger Dask workflow.

This PR adds two new params to `pixel_tides` which make Dask more configurable:
- `dask_compute`: Defaults to True; if set to False then data will be returned as a lazily loaded Dask array
- `dask_chunks`: Defaults to "auto"; this will automatically set Dask chunks to those in the input `ds` dataset (if they exist). Can also be used to pass in a tuple of custom chunks (e.g. `(2048, 2048)`) to use for the resampling step.

Also added a new test covering this functionality, and re-run the `How_to_guides/Tidal_modelling.ipynb` notebook.


vnewey (approved):
Thanks Robbi!

Issue #1090, opened 2023-06-29
Create new "DEA_datasets" notebook for Mangroves product 🌳


robbibt:

erialC-P:
Issue resolved :)

PR #1131, opened 2023-09-08
New DEA Mangroves notebook for inclusion with DEA products notebooks

erialC-P:
### Proposed changes
This is a new notebook for inclusion with the other DEA_products notebooks, introducing the DEA Mangroves dataset.

### Closes issues 
- Closes Issue #1090 - https://github.com/GeoscienceAustralia/dea-notebooks/issues/1090

- [-] Test notebook on both the `NCI` and `DEA Sandbox` (flag if not working as part of PR and ask for help to solve if needed)

This notebook has not been tested for use on the NCI.

Note: when plotting the timeseries statistics in this notebook, I opted to leave in some commented out code which refers to a mostly insignificant extra class in the dataset `not_observed`. Users can optionally include this class in their output plots if they like. However, I'm open to removing this from the notebook for simplicity.

Thanks for reviewing! :)


treefern:
I'll leave formal review for those more experienced with the relevant libraries. It looks great though!

robbibt (changes requested):
Hey @erialC-P, this is super awesome! Really really like it - the applications at the end are particularly neat and will definitely be useful for users trying to work out how to use the data.

I've left some comments via the notebook review app - mainly things to increase consistency with the other "DEA_products" notebooks. They have some slightly different standardised heading structures compared to other notebooks in the repo - it'd be nice to keep that going so the user experience of flicking between them is smooth. Could just copy the headings from here: https://docs.dea.ga.gov.au/notebooks/DEA_products/DEA_Water_Observations.html

I also don't think we need the uploaded "DEA_Mangroves.mp4" file; the video should already be embeded in the notebook.

Finally, one totally optional one: would it be worth reducing the spatial extent loaded by the notebook? We could probably zoom in a little closer to one area of mangroves, which might make data load quicker and some of the plots easier to see (with the current zoom you need to squint to see the different classes). Low priority though, it's fine as it is!

erialC-P:
Thanks for the review @robbibt! Super useful feedback and comments - I'm happy to accept them all!

robbibt:
Awesome - just realised this will need to be added to the `README.rst` file for it to render in the docs too.

And I don't think I hit commit  on deleting the MP4 file on my end - can you delete it in your branch? (just deleting the file then committing the deleted file should do it)

Then I think we're good to go!

robbibt (approved):
Looks good to me - ready to merge when you are!

Issue #1134, opened 2023-09-12
Allow `model_tides` to work for one timestep per one point


robbibt:
Currently `model_tides` automatically runs multiple timesteps for each tide modelling lat/lon point. This serves our main DEA use case: predicting tides at set tide modelling locations for every timestep in a satellite timeseries. This is effectively:

- One point, many timesteps

However, another common use case is to model tide heights where each lat/lon point has a unique timestep. E.g.:

- One point, one timestep

We should also support this second use case, perhaps by adding a new "mode" param:

- "one-to-many"
- "one-to-one"

robbibt:
Closed by #1142

PR #1142, opened 2023-09-20
Add new "one-to-one" mode for modelling tides at multiple locations with unique timesteps

robbibt:
### Proposed changes
Currently, the `model_tides` function operates in **"one-to-many"** mode, which will model tides for every requested timestep, at every requested location. For example, if we provided five x, y coordinates and five timesteps, the function would return:
```
5 locations * 5 timesteps = 25 modelled tides
```
However, often a user may want to model tides for a list of locations, each with a unique timesteps. 

This change adds a new **"one-to-one"** mode, allowing users to model tides for only these exact pairs of locations and times. Using the previous example (five x, y coordinates and five timesteps) in "one-to-one" mode:
```
5 timesteps at 5 locations = 5 modelled tides
```

Example from notebook:
image

Other changes:

- Updated `How_to_guides/Tidal_modelling.ipynb` to document new functionality
- Added tests for new functionality

### Closes issues (optional)
- Closes Issue #1134 


vnewey (approved):
Nice work Robbi!

PR #1149, opened 2023-11-13
Pull pixel_tide resampling into function for easier re-use

robbibt:
### Proposed changes
This is a simple refactor that modifies `dea_tools.coastal.pixel_tides` to pull out our current resampling code (used to resample low resolution modelled tides into the higher resolution of our satellite data) into a separate function. 

This will allow us to more easily re-use this functionality in external workflows (e.g. for resampling our ensemble tide models in DEA Intertidal).


erialC-P (approved):
Hey @robbibt, this looks good to go!! Approved! : )

PR #1152, opened 2023-11-22
Implement comprehensive GitHub Actions integration tests using prod DEA ODC database

omad:
### Proposed changes

This PR makes use of some AWS infrastructure I've setup to enable GitHub Actions to connect to the Production DEA ODC PostgreSQL database.

This will allow us to run automated tests of all our code and notebooks using real data, without having to maintain an extracted subset, or any other awkward workarounds.

It uses the existing [OIDC authentication to AWS](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services) to create a temporary network tunnel using [Basti](https://www.basti.app/).

### Additional updates from @robbibt (10/02/2024)

Added scheduled entire-repo integration testing:

- Break repo testing into discrete sections (Beginner's guide, DEA products etc) which are run in parallel using a matrix strategy. This reduces the 4+ test run time down to <1 hour 
- Test against both `latest` and `stable` Sandbox images
- Run entire test suit on a regular schedule using Github Actions `cron` capability (needs fine tuning

New tests run each PR:
- Test only modified notebooks + DEA Tools on each PR for faster testing

Minor fixes:
- Fixed several Jupyter Notebooks that were revealed as broken by the tests

image


robbibt:
Ready for review @benjimin: for the first time (and huge thanks to @omad) we now have comprehensive integration testing across the entire of DEA Notebooks!

For now I've put the scheduled tests on [an hourly cron job](https://github.com/GeoscienceAustralia/dea-notebooks/pull/1152/files#diff-b3467655c955d96edd47cbda20b6731cc2b2992fa3994077a3ff6ee1191cac37R12), which is obviously major overkill - I'll reduce it down to something more reasonable once we merge this PR and test it all works.

robbibt:
@JM-GA @emmaai Would also be great to get your review or a quick look at this: this new testing framework replaces the bespoke custom database we used to use in our tests with direct access to the prod database, allowing us to expand our tests to the entire DEA Notebooks repository (compared to just a hand picked selection of notebooks).

(you can ignore changes to `How_to_guides`, `Real_world_examples` and `Scalable_machine_learning`; they are fixes to broken notebooks that were identified by the tests)

benjimin:
I would have been inclined to unbundle unrelated changes (e.g. various notebook modifications versus totally new CI) but not a deal breaker.

I'm curious whether in future it might be cleaner to use native GitHub Actions support for running jobs in arbitrary containers, rather than invoking docker explicitly, but can explore that at another stage.

Keen to have this merged and see how it performs in practice.

robbibt:
> I would have been inclined to unbundle unrelated changes (e.g. various notebook modifications versus totally new CI) but not a deal breaker.

Yep, fair enough - I didn't originally intend to make those changes, but wanted to see if I could get the entire test suite to pass once they came up as failed notebooks.

> I'm curious whether in future it might be cleaner to use native GitHub Actions support for running jobs in arbitrary containers, rather than invoking docker explicitly, but can explore that at another stage.

Happy to take advice on this and very open to improving it in the future - I've just used Damien's code here for now (I did manage to remove our Docker Compose code here though, as we no longer need it for generating a temporary test database).

omad:
> I'm curious whether in future it might be cleaner to use native GitHub Actions support for running jobs in arbitrary containers, rather than invoking docker explicitly, but can explore that at another stage.

It might be possible, but wouldn't be trivial.

I think you'd probably want `basti` running within a GHA Service container, which would have to be responsible for doing the OIDC dance with AWS to get credentials. Although, I'm not even sure that's possible. At the least I think it'd need a custom image.

There's also a challenge with getting the image, since that's also using OIDC to authenticate with AWS, and the GHA workflow syntax expects username/password instead.

Or maybe you could keep it much the same, but run `basti` on the host, except that might not work because the code running in the container might not be able to access ports exposed on the host. Not sure.

robbibt (approved):
Merging now to test cron scheduling functionality - happy to revisit and make any further changes as subsequent PRs.

Thanks so much for your help @omad!

PR #1155, opened 2023-12-12
Update coastal.py

erialC-P:
### Proposed changes
Implements a minor change to the _pixel_tides_resample function in coastal.py to enable dataset handling when dask_chunks are no longer associated with the dataset following dask compute.


robbibt (approved):
Looks good to me, let's see if it passes tests and merge if it does!

You'll be able to grab the new DEA Tools version number here once it gets merged to develop (and then can update the `requirements.in` file in `dea-intertidal`): https://pypi.org/project/dea-tools/#history

PR #1161, opened 2023-12-14
Fixed several links after running link checker tools on the site

benji-glitsos-ga:
### Proposed changes
Fixed several links after running link checker tools on the site

### Closes issues (optional)
None


robbibt:
Hey @benjamin-glitsos, the reason for those relative links to scripts and notebooks is so that users can click through them on the DEA Sandbox and have the coresponding script open up to access the code. It's pretty important, particularly for notebooks that form a set that are run in order (e.g. "Click to launch Notebook 2 of 5" etc).

Is there any way we can continue to use relative links in the Notebooks and still have them load correctly in Knowledge Hub? That used to work in the original DEA Docs.

benji-glitsos-ga:
Hi Robbi, I'll try to find a solution. The problem is that the relative links weren't working when rendered on the Knowledge Hub site.

robbibt:
Thanks @benjamin-glitsos! I've just realised that most of the relative links to notebooks and scripts were actually removed in the previous PR, not this one... we'll probably need to revert those changes to restore them so they work correctly for our Sandbox users, so some way of dealing with relative links in KH would be ideal. 

I wonder if this could be dealt with via yet more redirects? e.g. updating the auto-resolved URL with one that points to the correct KH location...

robbibt (approved):
Don't want to hold up your work so approving for now, but noting that we'll probably need to revert back to relative links in DEA Notebooks so that they work correctly for our Sandbox users.

PR #1165, opened 2024-01-04
Fixed links and changed the way table of contents is handled

benji-glitsos-ga:
### Proposed changes
Fixed links and changed the way table of contents is handled. (I needed to change the 'toctrees' in order to add subheadings to the sidebar on the new DEA Knowledge Hub)

### Closes issues (optional)
None


Beginners_guide/01_Jupyter_notebooks.ipynb
robbibt, line 1: Just for this template file, I think our priority is for users to be able to click the link and automatically have the Jupyter notebook open on the Sandbox - even if that causes a broken link in Knowledge Hub.

Is it possible to leave this as a relative link, or will that cause issues on the KH side?

Beginners_guide/README.rst
```diff
@@ -3,32 +3,14 @@ Beginner's guide
 
 The Beginners Guide contains introductory notebooks aimed at introducing Jupyter Notebooks, Digital Earth Australia (DEA), and how to load, plot and interact with data from DEA. These notebooks are designed to be worked through in the following order:
 
-.. toctree::
```
robbibt, line 6: We used to use these readme tables of contents to control what notebooks are included in the docs (we have some notebooks that we sync to the Sandbox but don't render in the docs). Now that they've been removed, is there another place where we can do this, e.g. view and modify the notebooks that will be rendered in Knowledge Hub?

robbibt (changes requested):
Overall this looks fantastic - made one suggested change for the "template" link, and a question about how the tables of contents work now. 🙂

Beginners_guide/README.rst
benji-glitsos-ga: Hi Robbi, I will be keeping all of these `toctree` elements within this Notebooks repo except this one. That is because I am moving control of the top-level toctree to the dea-docs repo, as this allows more flexibility in how the sidebar displays on the DEA Knowledge Hub. However, I will explain it better when I make a new PR.

Beginners_guide/01_Jupyter_notebooks.ipynb
benji-glitsos-ga: I'll revert this change in the new PR I create. After deployment, I will figure out if there is a way to make it work on the new site, but the Sandbox is more important in this case.

PR #1170, opened 2024-01-08
Fixed only four links

benji-glitsos-ga:
### Proposed changes
This PR contains around 22 commits however most of these are commits that I have reverted or are commits that have already been merged to the develop branch. Therefore, you will notice that this PR only changes 4 links which are broken links on the DEA Knowledge Hub, but it maintains them as relative links so they will still work properly on the Sandbox.

### Closes issues (optional)
None


erialC-P (approved):
Looks good Benji! I think that maintaining the working links in the sandbox is the right call. Hopefully we can find a workaround for the docs :)

PR #1171, opened 2024-01-09
Add threads flag to parallel apply

alexgleith:
### Proposed changes
The `parallel_apply` function uses a process pool to do work in parallel, which is not necessarily
the best way to do parallel work. This PR keeps that as the default but also provides an option
to us threading as well.

### Closes issues (optional)
- n/a

- [] Ensure that all code cells follow the [PEP8 standard](https://www.python.org/dev/peps/pep-0008/) for code. The `jupyterlab_code_formatter` tool can be used to format code cells to a consistent style: select each code cell, then click `Edit` and then one of the `Apply X Formatter` options (`YAPF` or `Black` are recommended).

### DIFF
--- a/Tools/dea_tools/datahandling.py
+++ b/Tools/dea_tools/datahandling.py
@@ -929,6 +929,8 @@ def parallel_apply(ds, dim, func, use_threads=False, *args, **kwargs):
     use_threads : bool, optional
         Whether to use threads instead of processes for parallelisation.
         Defaults to False, which means it'll use multi-processing.
+        In brief, the difference between threads and processes is that threads
+        share memory, while processes have separate memory.
     *args :
         Any number of arguments that will be passed to `func`.
     **kwargs :
--- a/Tools/dea_tools/datahandling.py
+++ b/Tools/dea_tools/datahandling.py
@@ -20,28 +20,25 @@ on Github (https://github.com/GeoscienceAustralia/dea-notebooks/issues/new).
 Last modified: June 2023
 """
 
+import datetime
+
 # Import required packages
 import os
-import zipfile
-import datetime
-import requests
 import warnings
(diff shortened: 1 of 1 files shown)


BexDunn (changes requested):
Hey Alex, looks really handy! 
- Would a quick explainer on the differences fit into any of the existing notebooks?
- Can you please update the date modified up top?

alexgleith:
Added the last modified date.

I don't think this is used in Notebooks. It is used in Coastlines and other places.

From more testing, I'm not sure it makes a massive difference, but I think it's worth having as an option still.

robbibt:
This is fantastic @alexgleith! As someone from a non-ICT background I don't really have as good an understanding of this stuff as I should - any chance you could add one additional sentence to the doc string explaining why/when using threads vs processes might be a better option? (just something not super technical to help beginner users make a more informed choice)

alexgleith:
No worries, @robbibt. Done.

The longer explanation from my (still kind of lay person) view is that a process is a whole new operation, whereas a thread is running in the existing operation. Think of it as like spawning a new machine (process) to do some work, instead of doing another piece of work on the existing machine (thread).

A process is very much separated, and inter-process communication is hard, whereas threads share a single process, and so can share memory and communicate directly that way.

Being "thread safe" is important too, to stop tasks stomping on each other's memory... I think most of what this function is being used for will be thread safe. Most stuff in Python is thread safe, these days, but it does need consideration.

My $0.02!

robbibt (approved):

BexDunn (approved):
Thanks, good to have the extra explanation!

alexgleith:
Ok cool!

I can't merge it in. I have no power here 😆 

No rush. Maybe @omad has an opinion still.

PR #1172, opened 2024-01-09
Add course to USAGE.rst

BexDunn:
Add UNE course to courses list

### DIFF
--- a/.github/workflows/configs/spellcheck_wordlist.txt
+++ b/.github/workflows/configs/spellcheck_wordlist.txt
@@ -8,6 +8,7 @@ Almeida
 ANU
 apache
 ard
+automodule
 autosummary
 bandindices
 Biogeosciences
--- a/USAGE.rst
+++ b/USAGE.rst
@@ -55,6 +55,7 @@ Conferences
 
 Courses and training
 --------------------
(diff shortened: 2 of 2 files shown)


robbibt:
@BexDunn I can't work out why the spelling check fails here. It's complaining about "autosummary", but I can see it in the word list here: https://github.com/GeoscienceAustralia/dea-notebooks/blob/BexDunn-patch-2/.github/workflows/configs/spellcheck_wordlist.txt#L11

Any thoughts? If we can't work out what's going on, I'd be tempted to remove the spelling check - it's clunky and rigid and IMHO causes more problems than it solves...

BexDunn:
> @BexDunn I can't work out why the spelling check fails here. It's complaining about "autosummary", but I can see it in the word list here: https://github.com/GeoscienceAustralia/dea-notebooks/blob/BexDunn-patch-2/.github/workflows/configs/spellcheck_wordlist.txt#L11
> 
> Any thoughts? If we can't work out what's going on, I'd be tempted to remove the spelling check - it's clunky and rigid and IMHO causes more problems than it solves...

fixed it! 'automodule' was the problem this time. I see what you mean, maybe we can think about that one.

erialC-P (approved):
Approved! (Thanks for the tag - keeping me accountable! :) )

PR #1174, opened 2024-01-22
Remove spellcheck action and replace with PR checklist item

robbibt:
### Proposed changes
Hi all, we currently [have a spellcheck Github Action](https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/.github/workflows/spellcheck.yaml) to avoid spelling mistakes being committed into the repo. This Action uses [a large custom wordlist](https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/.github/workflows/configs/spellcheck_wordlist.txt) to check against, producing a failure if a word doesn't exist on the list.
 
In theory, this is a great idea. However, in practice it is often frustrating - the tool is highly inflexible and rigid, causing it to work poorly with acronyms, names and technical words, forcing users to continually update the wordlist. It has also recently been producing unexplained failures, which can be seen as failed checks on [a bunch of recent pull requests](https://github.com/GeoscienceAustralia/dea-notebooks/pulls).
 
We do actually already have a really nice (but little-known) built-in spelling tool on DEA Sandbox, that highlights misspelt words in pink and gives options for correcting them by right-clicking. IMO this is much more flexible and easier to use.
image
 
This PR:
- Removes the current Github Actions spelling workflow
- Updates the current Pull Request template checklist to add instructions to check for spelling mistakes using the existing spelling tool.


robbibt:
Example of the new PR checklist item:
image


https://github.com/GeoscienceAustralia/dea-notebooks/blob/spellchecker/.github/PULL_REQUEST_TEMPLATE.md

erialC-P (approved):
Makes sense and removes admin on the repo. Good idea!

PR #1177, opened 2024-01-28
Limit tide data downloaded in tests

robbibt:
### Proposed changes
As part of the DEA Intertidal project, we recently we updated the tide model data available on the S3 `dea-non-public-data` bucket to include data from an additional ~5 tide models. 

To avoid this new data being synced during the DEA Notebooks integration tests (potentially leading to longer run times), this PR updates the "Copy tide modelling files with the AWS CLI" step of our tests to only download a smaller subset of the tide model data.


CEKrause (approved):
Nice :)

PR #1182, opened 2024-02-12
Fixed a strange number appearing in the heading of this article on th…

benji-glitsos-ga:
…e KH

### Proposed changes

On the KH, the heading of this notebook was displaying with this strange number in it. Minor change: This change only adds one space character to the code to make the code the same as the other pages.

ksnip_20240212-125636


robbibt (approved):
Thanks so much, this has been bothering me for ages!

PR #1183, opened 2024-02-12
Fixing some bullet points displaying incorrectly on the KH

benji-glitsos-ga:
### Proposed changes

I added a newline above the bullet list on two pages because this is a Markdown formatting error that caused the bullet lists to render incorrectly on the knowledge hub. It does not change the way it displays on the Sandbox.

### Checklist 

I tested it in the DEA Sandbox.


robbibt (approved):
Great - this is a known issue that shows up from time to time. Good to get them fixed up!

PR #1187, opened 2024-02-18
Add user-friendly landing page to DEA Sandbox

robbibt:
### Proposed changes
This PR provides a user friendly landing page that will make it easier for users to learn about the content available in the Sandbox, and useful reference information in the Knowledge Hub:

image

### Closes issues (optional)
- Closes Issue #1188 

### Checklist 

sandbox_spellchecker


DEA_Sandbox.ipynb
geoscience-aman, line 1: Suggest `` maybe so it covers the whole width of a larger monitor screen without affecting the display on a smaller laptop screen?

geoscience-aman: When referring to DEA Tools, maybe we can link to the PyPI package? 
https://pypi.org/project/dea-tools/0.3.0/

geoscience-aman (approved):
Looks nice!

DEA_Sandbox.ipynb
benji-glitsos-ga: Hi @geoscience-aman , I think it would be better if the header image was wider but I tried applying a larger width and it stretches the image beyond its natural limit and makes it blurry unfortunately. Unless we have a larger version of dea_logo_wide.jpg somewhere @robbibt  ?

Also, it will be great if we link to the Pypy package however I think it would be better for the user experience if we put it on the DEA Tools cover page: https://docs.dea.ga.gov.au/notebooks/Tools/

I will create a PR to suggest adding it to that page.

robbibt: Yeah, I reckon we add a PyPI link somewhere here, so it's easily accessible to users accessing the Python API docs:
image

robbibt: I agree 💯  about the banner, but as far as I know it's the widest version we have... although I can't remember where it originally came from or if it was something custom we put together for DEA Notebooks.

Might need to be a future improvement for this one I think

BexDunn (approved):
I'm so excited about this :) Looks good

PR #1194, opened 2024-02-27
Update tests to ignore Markdown, RST and template and landing page notebooks

robbibt:
### Proposed changes
This PR updates the "ignore-paths" line of our Github Actions testing workflow to ignore future updates to the repo's Readme, template notebook or landing page notebook.

This will prevent us having to wait ~30 minutes for tests to run after making minor updates to Markdown or RST files.

### DIFF
--- a/.github/workflows/test_notebooks.yml
+++ b/.github/workflows/test_notebooks.yml
@@ -5,21 +5,19 @@ on:
   push:
     branches: [ develop, stable, nbtests ]
     paths-ignore:
-      - 'README.md' # ignore readme
       - 'DEA_Sandbox.ipynb' # ignore landing page
       - 'DEA_notebooks_template.ipynb' # ignore template
-      - '**/*.md' # ignore markdown files in nested folders
-      - '**/*.rst' # ignore restructured text files
+      - '*.md' # ignore all markdown files
+      - '*.rst' # ignore all restructured text files
       - '.github/**' # ignore anything in .github folder
       - '!.github/workflows/test_notebooks.yml' # except test_notebooks.yml
   pull_request:
     branches: [ develop, stable ]
     paths-ignore:
-      - 'README.md'
       - 'DEA_Sandbox.ipynb'
       - 'DEA_notebooks_template.ipynb'
-      - '**/*.md'
-      - '**/*.rst'
+      - '*.md'
+      - '*.rst'
       - '.github/**'
       - '!.github/workflows/test_notebooks.yml'
 
--- a/.github/workflows/test_notebooks.yml
+++ b/.github/workflows/test_notebooks.yml
@@ -5,13 +5,19 @@ on:
   push:
     branches: [ develop, stable, nbtests ]
     paths-ignore:
-      - '**/*.md' # ignore markdown files
+      - 'README.md' # ignore readme
+      - 'DEA_Sandbox.ipynb' # ignore landing page
+      - 'DEA_notebooks_template.ipynb' # ignore template
+      - '**/*.md' # ignore markdown files in nested folders
       - '**/*.rst' # ignore restructured text files
       - '.github/**' # ignore anything in .github folder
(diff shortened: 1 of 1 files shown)


.github/workflows/test_notebooks.yml
```diff
@@ -5,15 +5,19 @@ on:
   push:
     branches: [ develop, stable, nbtests ]
     paths-ignore:
-      - '**/*.md' # ignore markdown files
-      - '**/*.rst' # ignore restructured text files
+      - 'DEA_Sandbox.ipynb' # ignore landing page
+      - 'DEA_notebooks_template.ipynb' # ignore template
+      - '*.md' # ignore all markdown files
+      - '*.rst' # ignore all restructured text files
```
benji-glitsos-ga, line 11: I think that this `**/*` syntax is needed to ignore all of them recursively, not just in the root directory.

```yaml
      - '**/*.md' # ignore markdown files
      - '**/*.rst' # ignore restructured text files
```

.github/workflows/test_notebooks.yml
```diff
@@ -5,15 +5,19 @@ on:
...
       - '.github/**' # ignore anything in .github folder
       - '!.github/workflows/test_notebooks.yml' # except test_notebooks.yml
   pull_request:
     branches: [ develop, stable ]
     paths-ignore:
-      - '**/*.md'
-      - '**/*.rst'
+      - 'DEA_Sandbox.ipynb'
```
benji-glitsos-ga, line 17: Are you sure this should be ignored? It is a Notebook file that we need to check renders properly on the Sandbox.

.github/workflows/test_notebooks.yml
```diff
@@ -5,15 +5,19 @@ on:
...
+      - 'DEA_notebooks_template.ipynb'
+      - '*.md'
+      - '*.rst'
       - '.github/**'
       - '!.github/workflows/test_notebooks.yml'
```
benji-glitsos-ga, line 22: Maybe we should ignore everything in the root directory except for Notebook files:

```yaml
      - '*'
      - '!*.ipynb'
```

.github/workflows/test_notebooks.yml
robbibt: It doesn't actually include any code - it's all entirely pre-rendered markdown cells so there's not anything for `pytest` to run to test for a pass/fail.

.github/workflows/test_notebooks.yml
robbibt: I'd be happy to ignore all top-level files - the two notebooks there are either entirely Markdown or are not intended to be run by our users (both [aren't included in our test coverage anyway](https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Tests/test_notebooks.sh#L14), and `DEA_notebooks_template.ipynb` [isn't even synced to the Sandbox](https://github.com/GeoscienceAustralia/dea-sandbox/blob/develop/docker/assets/sync_repo#L15)).

.github/workflows/test_notebooks.yml
robbibt: I don't think the recursive ignore was ignoring the top level markdown files in your previous PR - it seemed to be only catching the ones nested within subfolders.

We could leave the recursive ignore as it is though and adding in your suggested top level "*" ignore to catch the Readme?

robbibt: Actually, maybe it just needs to be "**.md?" to catch both top level and nested files?
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths

.github/workflows/test_notebooks.yml
benji-glitsos-ga: I agree. Sounds good.

BexDunn (approved):

robbibt:
Superceded by #1152

PR #1205, opened 2024-03-12
Add updated xr_interpolate func and tests

robbibt:
### Proposed changes
This PR updates the existing `interpolate_2d` function into a brand new `xr_interpolation` function. Compared to `interpolate_2d`, this function:

- Supports a `geopandas.GeoDataFrame` points dataset as an input
- Automatically interpolates all numeric columns in this dataset 
- Also supports Inverse Distance Weighted interpolation, with option to restrict the interpolation to `k` nearest neighbours.

Example Inverse Distance Weighted interpolated  `xarray` output for a point `gdf` containing 7 columns:

image

Have added comprehensive tests, and deprecated `interpolate_2d` to be removed at a later date.

### Checklist 

- [ ] Include relevant tags in the final notebook cell (refer to the [DEA Tags Index](https://knowledge.dea.ga.gov.au/genindex/), and re-use tags if possible)

sandbox_spellchecker

### DIFF
--- a/Tests/dea_tools/test_spatial.py
+++ b/Tests/dea_tools/test_spatial.py
@@ -8,7 +8,13 @@ import geopandas as gpd
 import datacube
 from datacube.utils.masking import mask_invalid_data
 
-from dea_tools.spatial import subpixel_contours, xr_vectorize, xr_rasterize
+from dea_tools.spatial import (
+    subpixel_contours,
+    xr_vectorize,
+    xr_rasterize,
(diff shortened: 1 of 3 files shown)


erialC-P (commented):
Looks good to me. Your error reporting in spatial.py is a masterclass! I think that it will pick up a huge number of the likely errors as users use that script :)

erialC-P (approved):

PR #1211, opened 2024-03-18
Update R^2 method in `eval_metrics`

robbibt:
### Proposed changes
The current R^2 coefficient of determination method in `eval_metrics` produces confusing and unpredictable results. This PR simplifies this by simply squaring the linear regression correlation co-efficient, as recommended by the `stats.linregress` function:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.linregress.html
image


vnewey (approved):
Good idea

PR #1213, opened 2024-04-03
Computing WO from S2 ARD

Yukienip:
### Proposed changes
New notebook to demonstrate applying WO classifier to S2 imagery. Provides options for x-y input and custom bounding box. Chosen area for demonstration is affected by cloud masking error, and solutions are explored in the following notebook (Name TBC, currently Incorporating_Fmask_ARD)

### Closes issues (optional)
- Closes Issue #000

### Checklist 

- [x ] Remove any unused/empty code cells
- [ x] Include relevant tags in the final notebook cell (refer to the [DEA Tags Index](https://knowledge.dea.ga.gov.au/genindex/), and re-use tags if possible)
- [ x] Clear all outputs, run notebook from start to finish, and save the notebook in the state where all cells have been sequentially evaluated
- [sandbox only ] Test notebook on both the `NCI` and `DEA Sandbox` (flag if not working as part of PR and ask for help to solve if needed)
- [ x] If applicable, update the `Notebook currently compatible with the NCI|DEA Sandbox environment only` line below the notebook title to reflect the environments the notebook is compatible with
- [x ] Check for any spelling mistakes using the DEA Sandbox's built-in spellchecker (double click on markdown cells then right-click on pink highlighted words). For example:

sandbox_spellchecker


BexDunn:
Hey @Yukienip do you have more files in here than the ones you have updated? you may want to look at doing this one from a fresh branch and pulling your updates on a fresh checkout of develop?

robbibt (changes requested):
Hey @Yukienip! This is an incredible piece of work, and something we definitely want to preserve and make accessible for future reference. 🚀 Thinking about the range of content this includes, I do think that the public-facing `develop`/`stable` branches of DEA Notebooks may not be quite the right place for it as these are primarily for self-contained tutorial-style content: this is more of a research output with supplementary data and sample outputs (similar to some material we used to include in the `Scientific_workflows` folder in this repo that we removed from `develop` last year).

I think this content is better suited to stay on a discrete branch (not merged into `develop`) where it will still be accessible and available, but sit as an independent set of content. So it doesn't get lost, we can set up some branch protection rules to lock it down, and also have a think about a nice place where we can link to this material so it's also discoverable. There's some precedents for this kind of "research"-style content on the branches below:

- https://github.com/GeoscienceAustralia/dea-notebooks/tree/scientific_workflows_archive
- https://github.com/GeoscienceAustralia/dea-notebooks/tree/subpixel_waterlines
- https://github.com/GeoscienceAustralia/dea-notebooks/tree/coastal-waterbodies

Let me know what you think! (Another option could be to try and simplify the content down to a single stand-alone notebook, but I'd worry we'd lose too much useful content if we did that)

PR #1219, opened 2024-04-10
Point DEA Intertidal KH link to product page, not notebook

robbibt:
### Proposed changes
The current Knowledge Hub link accidently points to the DEA Notebooks product notebook page, not the product page itself. This fixes it.


vnewey (approved):
Thanks Robbi!

PR #1228, opened 2024-05-22
Deleted the 'gen' folder used by Sphinx autosummary

benji-glitsos-ga:
### Proposed changes
Deleted this folder because it is generated at build time by Sphinx autosummary plugin and therefore, it does not need to be version controlled.


BexDunn (approved):
Seems sensible :)

PR #1229, opened 2024-05-22
Moved the mock_imports file into the dea-notebooks repo

benji-glitsos-ga:
The mock_imports.txt file was previously in the dea-knowledge-hub repo, but that was inconvenient since actually it relates to the DEA Tools pages, so I moved it into this repo so that it will be easier to edit in future.


BexDunn (approved):
Looks good - worth mentioning somewhere in the readme for dea-notebooks what this one is for too?

robbibt (approved):

benji-glitsos-ga:
I have put on my todo list to write a Wiki article for how to create/update these DEA Tools Python modules since we don't currently have a Wiki article on this.

PR #1240, opened 2024-07-11
reorganize dask compute to avoid repeated load in wetland turbidity notebook

uchchwhash:
### Proposed changes
Invoke `dask` compute at the opportune moment. Saves about 9mins of runtime for me. Uses about 500MB more memory.

### Checklist 


robbibt (approved):
Awesome!

PR #1244, opened 2024-07-11
Speed up WIT notebook

BexDunn:
### Proposed changes
- Reduced the time period for execution from two years to one year
- Added and documented the ability to have months on the graphs instead of years
- Removed a hard-coded infer_datetime parameter, as pandas doesn't need it anymore and warns if you include it
- added explicit dask chunks

This could probably still go faster, I'm not sure why I'm loading into memory so early here. But this should make a big difference for testing in the meantime


robbibt (changes requested):
Hey @BexDunn, this looks great. For the Dask chunking, I think you can set the x and y chunks to something bigger and it will still have the same effect (Dask will clip the chunks back to the size of your data): e.g. `"x": 2048, "y": 2048`. That way it will be more generic to other areas/extents in the future (but hopefully still keep the spatial chunking you've set here for this example).

robbibt (approved):
Looks great!

PR #1269, opened 2024-09-23
Minor compatibility change for tide modelling package

robbibt:
### Proposed changes
Minor change to pin the pyTMD tide modelling package to a set of compatible versions that work with DEA Tools.


erialC-P (approved):
Approved :)

PR #1270, opened 2024-09-30
Mitch predict_xr

mitchest:
### Proposed changes
Updating the `predict_xr()` function to allow returning the full array of prediciton probabilities our of `predict_proba()`, which is actually the default behaviour normally

### Closes issues (optional)
- Closes Issue #000

### Checklist 

- [x ] Remove any unused Python packages from `Load packages`
- [ x] Remove any unused/empty code cells
- [x ] Include relevant tags in the final notebook cell (refer to the [DEA Tags Index](https://knowledge.dea.ga.gov.au/genindex/), and re-use tags if possible)
- [x ] Clear all outputs, run notebook from start to finish, and save the notebook in the state where all cells have been sequentially evaluated
- [x ] Check for any spelling mistakes using the DEA Sandbox's built-in spellchecker (double click on markdown cells then right-click on pink highlighted words). For example:

sandbox_spellchecker

### DIFF
--- a/Tools/dea_tools/classification.py
+++ b/Tools/dea_tools/classification.py
@@ -336,7 +336,7 @@ def predict_xr(
             print("   probabilities...")
             out_proba = model.predict_proba(input_data_flattened)
 
-            # convert to %
+            # return either one band with the max probability, or the whole probability array
             if max_proba == True:
                 print("  returning single probability band.")
                 out_proba = da.max(out_proba, axis=1) * 100.0
@@ -348,12 +348,10 @@ def predict_xr(
             else:
                 print("  returning class probability array.")
                 out_proba = out_proba * 100.0
-                
                 class_names = model.classes_  # Get the unique class names from the fitted classifier
 
-                probabilities_dataset = xr.Dataset()
-
                 # Loop through each class (band)
+                probabilities_dataset = xr.Dataset()
                 for i, class_name in enumerate(class_names):
                     reshaped_band = out_proba[:, i].reshape(len(y), len(x))
                     reshaped_da = xr.DataArray(
--- a/Tools/dea_tools/classification.py
+++ b/Tools/dea_tools/classification.py
@@ -288,7 +288,7 @@ def predict_xr(
             input_xr.chunks["y"][0]
         )
 
-    def _predict_func(model, input_xr, persist, proba, proba_max, clean, return_input):
+    def _predict_func(model, input_xr, persist, proba, max_proba, clean, return_input):
         x, y, crs = input_xr.x, input_xr.y, input_xr.geobox.crs
 
         input_data = []
@@ -337,7 +337,7 @@ def predict_xr(
             out_proba = model.predict_proba(input_data_flattened)
 
             # convert to %
-            if proba_max == True:
+            if max_proba == True:
                 print("  returning single probability band.")
                 out_proba = da.max(out_proba, axis=1) * 100.0
             else:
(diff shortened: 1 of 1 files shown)


robbibt:
Hey @mitchest, this looks great to me - do you have a screenshot of the outputs of `proba_max=True` and `proba_max=False` that you could share here just to clarify what results you get using each option?

mitchest:
> Hey @mitchest, this looks great to me - do you have a screenshot of the outputs of `proba_max=True` and `proba_max=False` that you could share here just to clarify what results you get using each option?

Yer - no worries

Here's a SS from just the outputs (plot 1/3 is `proba_max=False`, plot 2 is `proba_max=True`)
image

Here's a SS actually using the coastal binary layer to mask another product
image

robbibt (approved):

jessjaco:
I noticed reduced performance after this was merged. After reviewing the code it looks like these two lines no longer have an effect, and that `clean=True` is perhaps broken. I think they need to be moved back up where they were.

https://github.com/GeoscienceAustralia/dea-notebooks/blob/a6e937f89fd24ea91a6abcaf42e8f1dd09cb0c85/Tools/dea_tools/classification.py#L365-L366

robbibt:
@jessjaco By performance, do you mean speed is slower, or that the results are lower quality? Any extra info (e.g. screenshots or timings) would be awesome - we'll have a look!

Are you running your code with `max_proba=True` and `clean=True`?

jessjaco:
It is slower, and the results are not masked as they should be. I think the speed issue is due to the fact it is predicting over data that should be masked. If you look at the current code you can see the section I quoted above can't have an effect (since `out_proba` has already been reassigned to `output_xr["Predictions"]`). See below for a reproducible demo. I can make it a separate issue if you want (and try a pull request), this is just the change that caused the problem.

```python
from dea_tools.classification import predict_xr
import numpy as np
import odc.geo.xr
from sklearn.linear_model import LinearRegression
import xarray as xr

# Dummy data & model
X = np.random.rand(1000, 1)
B = 3
e = np.random.rand(1000, 1) / 4

y = X * B + e

model = LinearRegression()
model.fit(X, y)

size_1d = 1_000
X_new = np.random.rand(size_1d, size_1d)
X_new_ds = (
    xr.DataArray(
        X_new,
        dims=("x", "y"),
        coords={"x": np.arange(size_1d), "y": np.arange(size_1d)},
    )
    .odc.assign_crs(26912)
    .to_dataset(name="data")
)

prediction = predict_xr(model, X_new_ds, chunk_size=10)
print(prediction.Predictions.isnull().any())
# False and should be

X_new_ds_masked = X_new_ds.where(X_new_ds < 0.9)
masked_prediction = predict_xr(model, X_new_ds_masked, chunk_size=10, clean=True)

print(X_new_ds_masked.data.isnull().any())
# True and should be

print(masked_prediction.Predictions.isnull().any())
# False, but should be True
```

jessjaco:
I was noticing performance issues, and I'm not positive why. But the results also appear to be incorrect.  Please see below.

```python
from dea_tools.classification import predict_xr
import numpy as np
import odc.geo.xr

from sklearn.tree import DecisionTreeClassifier
import xarray as xr

# Dummy data & model
X = np.random.rand(1000, 1)
B = 3
e = np.random.rand(1000, 1)

y = ((X * B + e) // 1).astype(int) + 1

model = DecisionTreeClassifier()
model.fit(X, y)

size_1d = 1_000
X_new = np.random.rand(size_1d, size_1d)
X_new_ds = (
    xr.DataArray(
        X_new,
        dims=("x", "y"),
        coords={"x": np.arange(size_1d), "y": np.arange(size_1d)},
    )
    .odc.assign_crs(26912)
    .to_dataset(name="data")
)

prediction = predict_xr(model, X_new_ds, chunk_size=10, proba=True)
print((prediction.Probabilities == 0).any())
# False

X_new_ds_masked = X_new_ds.where(X_new_ds < 0.9)
masked_prediction = predict_xr(
    model, X_new_ds_masked, chunk_size=10, proba=True, clean=True
)

print(X_new_ds_masked.data.isnull().any())
# True and should be

print((masked_prediction.Probabilities == 0).any())
# False, but should be true, since the nans should be converted to zero
```

Issue #1271, opened 2024-10-02
Scalable Machine Learning workflow - fails 


gflod:
**Describe the bug/issue**
The notebook seems to be calling on the incorrect bands for the product being called

**Steps to reproduce**
Steps to reproduce the behaviour. 
1. Open DEA Sandbox
2. Launch the "Real_world_examples/Scalable_machine_learning/4_Classify_satellite_data.ipynb" notebook
3. Run all cells
4. Code in section "Loop through test locations and predict" produces error "--------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
File ~/Real_world_examples/Scalable_machine_learning/../../Tools/dea_tools/bandindices.py:410, in calculate_indices(ds, index, collection, custom_varname, normalise, drop, inplace)
    409     mult = 10000.0 if normalise else 1.0
--> 410     index_array = index_func(ds.rename(bands_to_rename) / mult)
    411 except AttributeError:

File ~/Real_world_examples/Scalable_machine_learning/../../Tools/dea_tools/bandindices.py:152, in calculate_indices.<locals>.<lambda>(ds)
    149 # Dictionary containing remote sensing index band recipes
    150 index_dict = {
    151               # Normalised Difference Vegation Index, Rouse 1973
--> 152               'NDVI': lambda ds: (ds.nir - ds.red) /
    153                                  (ds.nir + ds.red),
    154     
    155               # Non-linear Normalised Difference Vegation Index,
    156               # Camps-Valls et al. 2021
    157               'kNDVI': lambda ds: np.tanh(((ds.nir - ds.red) /
    158                                            (ds.nir + ds.red)) ** 2),
    159 
    160               # Enhanced Vegetation Index, Huete 2002
    161               'EVI': lambda ds: ((2.5 * (ds.nir - ds.red)) /
    162                                  (ds.nir + 6 * ds.red -
    163                                   7.5 * ds.blue + 1)),
    164 
    165               # Leaf Area Index, Boegh 2002
    166               'LAI': lambda ds: (3.618 * ((2.5 * (ds.nir - ds.red)) /
    167                                  (ds.nir + 6 * ds.red -
    168                                   7.5 * ds.blue + 1)) - 0.118),
    169 
    170               # Soil Adjusted Vegetation Index, Huete 1988
    171               'SAVI': lambda ds: ((1.5 * (ds.nir - ds.red)) /
    172                                   (ds.nir + ds.red + 0.5)),
    173   
[...]

PR #1277, opened 2025-01-16
small fixes in input for calculate_indices() rgb()

GL-S:
### Proposed changes
added line to prevent warnings from dask.
Changed collection name in calculate_indices(), and reset band names in rgb() to default to avoid errors

### Closes issues (optional)
- Closes Issue #1271

### Checklist 


Beginners_guide/01_Jupyter_notebooks.ipynb
```diff
@@ -60,7 +60,7 @@
     "These options will run a single cell at a time.\n",
     "\n",
     "To automatically run all cells in a notebook, navigate to the \"Run\" tab of the menu bar at the top of JupyterLab and select \"Run All Cells\" (or the option that best suits your needs).\n",
-    "When a cell is run, the cell's content is executed.\n",
+    "When a cell is run,the cell's content is executed.\n",
```
JM-GA: @GL-S - might need to revert this change - guessing it was the one testing git access

GL-S: Yes you are correct, I will

robbibt:
This looks great, thankls @GL-S! I left a few minor comments on the notebook... for the Dask thing, my main priority is just making sure the new line doesn't accidently cause any other downstream issues. If we're sure it doesn't, then we could consider moving it from this notebook to the `dea_tools.dask` module instead to hide it from external users?

GL-S:
Thank you for the comments @robbibt . 
The warnings appeared when running the model for classifying the satellite images, there is a lot of printing which doesn't look great.
 I added the line about the dask's warnings because it was a solution suggested by the warning itself (screenshot added below). I am actually not sure what it does in details, it should have an influence on how the processes run and exit. It's probably safer not to add it to `dea_tools.dask` yet. 
 ...maybe a `filterwarnings` in  `dea_tools.dask` is a better solution? It doesn't work when added to the 4_Classify_satellite_data.ipynb

warnings dask

Regarding the comments about the band names, I get your point. I will remove those commented lines

robbibt:
Thanks @GL-S! Yeah, those warnings look terrible, it would definitely be nice to hide them away.

@emmaai Do you know what these Dask warnings in the screenshot above mean, and whether adding `dask.config.set({'distributed.worker.daemon': False})` to our `dea_tools.dask` script would be a safe move?

emmaai:
The warning seems to suggest there was a conflict between config worker as daemon and number of jobs = 1. Intuitively I don’t know why it can’t be both. The suggested solution is setting worker process as non-daemon. I don’t think it’s a good idea in general. Also as it runs with scikit-learn, I suspected the warnings actually suggested something different and more serious. Scikit-learn is multi-threading, it’s tricky to run with dask. I didn’t read the notebook. If it couldn’t be understood and resolved with reasonable thought, I’d suggest to keep it until further investigation.

GL-S:
Hi all, I have submitted another commit.
I removed the `dask.config.set({'distributed.worker.daemon': False})` to be safe, as Emma mentioned that it might not be a good idea. I will follow up on this when the notebooks COP resumes, as James suggested.
I have also removed the commented lines about the default band names.
Effectively, the changes from the version in develop are the updated product and collection names in the function `feature_layers`. The warnings are still there, but at least the notebook runs with no errors.

robbibt (approved):
Thanks heaps @GL-S, I think this is a good approach for now. Any chance you could create quick Github issue in this repo describing the warnings issue, that we can refer to later to work out what's going wrong?

Other than that, I think this is ready to "Squash and merge" into the develop branch 🙂

GL-S:
Sure, I will create an issue now

PR #1284, opened 2025-02-19
Loosen `pyTMD` version restriction, fix broken DEA Tools build

robbibt:
### Proposed changes
Removes the upper limit on the `pyTMD` tide modelling package in favour of excluding a specific version. This will allow DEA Tools to be used more easily in our external DEA Coastlines code which requires a higher version of `pyTMD`.

This PR also includes an attempt to fix the failing DEA Tools package build here, by pinning Python to <3.12: https://github.com/GeoscienceAustralia/dea-notebooks/actions/runs/13402753502

This issue may have been caused by a recent update to `ubuntu-latest`, which uses Python 3.12 by default (which is incompatible with the older `pip==22.0.4` version we use to build the package).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,10 +16,12 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - uses: actions/checkout@v2
+      
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.11'  # Pin to <3.12 for compatibility with pip==22.0.4
+          python-version: '3.11'  # Specify a stable Python version
       
       - name: Get history and tags for SCM versioning
         run: |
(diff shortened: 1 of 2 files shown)


BexDunn (approved):
Happy to approve to go ahead if that helps. Will we need an issue to track the fact that we have pinned the tools build, such that we can remember to unpin it when our other packages catch up?

robbibt:
> Happy to approve to go ahead if that helps. Will we need an issue to track the fact that we have pinned the tools build, such that we can remember to unpin it when our other packages catch up?

I'm going to see if this fixes it and revert quickly if not. 🙂 

As far as I can tell, there's not a big drawback to pinning the Python versions/packaging tools... it's just used for this specific Github Action, and prevents things from changing in the background when Github changes their Github Actions defaults. But eventually this package build set up will become pretty ancient and we might want to move to a more modern/updated approach.

BexDunn:
Nice! Carry on then :) Thanks for fixing it!

PR #1299, opened 2025-02-27
Update test suit to fix failures and improve setup

robbibt:
### Proposed changes
Tide modelling functions in in `Tools/dea_tools/coastal.py` have been deprecated and moved to [eo-tides](https://geoscienceaustralia.github.io/eo-tides/), where they are now extensively tested (e.g. https://github.com/GeoscienceAustralia/eo-tides/tree/main/tests). Because of this, we can simplify our DEA Notebooks tests by removing these tests from the repo.

In addition, have updated our tests so that:

- We no longer run DEA Tools tests on every PR. Instead, only modified notebooks are tested. This should make most PR tests much faster.
- DEA Tools tests (as well as entire repo Notebooks tests) are now only run if DEA Tools Python files are modified. We need to test the entire repo because changes to DEA Tools could possibly affect any part of the repo.


robbibt:
These test failures are OK/expected - they indicate that our code will fail if run on the new experimental "latest" Sandbox image that @emmaai is working on. So we will need to fix some things before that image is eventually rolled out to the Sandbox - but that's months away! 🙂 

image

erialC-P (approved):
Thanks for this work Robbi! Awesome changes to the repo testing schedule!!

Issue #1267, opened 2024-09-20
Platform change for ERA5 climate data affects External_data_ERA5_Climate.ipynb


debs21:
**Describe the bug/issue**
The API for downloading ERA climate data has recently altered - former method no longer supported.

**Steps to reproduce**
Here are some links that describe the changes:
https://confluence.ecmwf.int/display/CKB/Please+read%3A+CDS+and+ADS+migrating+to+new+infrastructure%3A+Common+Data+Store+%28CDS%29+Engine
https://cds-beta.climate.copernicus.eu/how-to-api
https://forum.ecmwf.int/t/informative-page-in-readiness-for-new-cds-beta/2985

**Expected behaviour**
Update API call.

**Environment information**
(https://app.sandbox.dea.ga.gov.au/user/dm939976@gmail.com/lab/tree/How_to_guides/External_data_ERA5_Climate.ipynb)

robbibt:
Thanks @debs21! We do need to update that notebook. For now, you can try out this experimental new notebook which uses a more cloud-friendly way of accessing the data (from Google's ARCO-ERA5 dataset):
https://github.com/GeoscienceAustralia/dea-notebooks/blob/rbt/Testing/ERA5_climate_data.ipynb

debs21:
Hi Robbi,

I’ll try it out today.

Thankyou
Deb

cbur24:
Complete #1335

PR #1335, opened 2025-04-08
New climate data notebook

cbur24:
### Proposed changes
The ERA5 climate notebook that is currently in the repo references data that has long since been depreciated. This PR removes the old ERA5 notebook and modules, and replaces it with a new notebook that shows how to stream Australian-specific climate data from the NCI using THREDDS, and how to stream global climate reanalysis data using Google's 'analysis ready' ERA5 collection.

### Closes issues (optional)
This PR address several 'cards' in the DEA-Notebooks community project. Related to #1267

### Checklist

- [X] Checked the structure of the notebook follows our [DEA-notebooks template](https://github.com/GeoscienceAustralia/dea-notebooks/wiki/NotebookTemplate)
- [X] Removed any unused Python packages from `Load packages`
- [X] Removed any unused/empty code cells
- [X] Removed any guidance cells (e.g. `General advice`)
- [ ] Ensured that all code cells follow the [PEP8 standard](https://www.python.org/dev/peps/pep-0008/) for code. The `jupyterlab_code_formatter` tool can be used to format code cells to a consistent style: select each code cell, then click `Edit` and then one of the `Apply X Formatter` options (`YAPF` or `Black` are recommended).
- [X] Included relevant tags in the final notebook cell (refer to the [DEA Tags Index](https://knowledge.dea.ga.gov.au/genindex/), and re-use tags if possible)
- [X] Tested notebook on the [DEA Sandbox](https://knowledge.dea.ga.gov.au/guides/setup/Sandbox/sandbox/)
- [X] Cleared all outputs, run notebook from start to finish, and save the notebook in the state where all cells have been sequentially evaluated
- [X] If applicable, update the `Notebook currently compatible with` line below the notebook title to reflect the environments the notebook is compatible with
- [X] Check for any spelling mistakes using the DEA Sandbox's built-in spellchecker (double click on markdown cells then right-click on pink highlighted words). For example:

sandbox_spellchecker


GL-S:
This is a useful notebook!

I have modified the part where the THREDDS URLs are retrieved.

I was not able to run the resampling to annual sum on a 2-years datacube because of memory errors. So the notebook now takes only one year as input. 

However, I have also provided an alternative notebook that still uses 2 years, but resamples them one at the time when plotting the maps. I have included it so you could have a look, but feel free to remove it if you prefer the one-year notebook

GL-S (approved):

Issue #1337, opened 2025-04-11
Implement contiguity `fuse_func` inside `load_ard`


robbibt:
Contiguity masking in `load_ard` is affected by a bug caused by continuity data not being properly combined during `group_by="solar_day"` operations. We can resolve this by using a custom `fuse_func` to ensure that valid 1 values are correctly returned for overlapping contiguity dataset outputs.

```
def _contiguity_fuser(dst: np.ndarray, src: np.ndarray) -> None:
    """
    Ensure contiguity data is properly combined by replacing
    pixels in `dst` that are either 0 (non-contiguous) or 255
    (nodata) with the corresponding value from `src`, propogating
    1 (valid contiguous data) if it exists.
    """
    np.copyto(dst, src, where=np.isin(dst, (255, 0)))
```

DEA Intertidal implementation:
https://github.com/GeoscienceAustralia/dea-intertidal/pull/119

PR #1341, opened 2025-04-24
Fix contiguity masking in `load_ard` by applying fuse function

robbibt:
### Proposed changes
Contiguity is a band in our ARD that identifies whether a pixel was missing data in _any_ of our surface reflectance bands (e.g. a pixel that is missing data in at least one band is considered "non-contiguous"). Masking by contiguity (`mask_contiguity=True` in `load_ard`) is important for applications like Geomedian, where missing data from individual bands can result in a biased output (particularly when some timesteps might be missing data in one band but others do not).

A bug in `load_ard` (#1337) however meant that contiguity band data was being combined incorrectly when multiple timesteps where merged using `group_by="solar_day"`, due to contiguity data from neighbouring images is pasted semi-randomly over each other. This produced nasty wedge-shaped artefacts like this:

image

This PR fixes this bug by adding a custom "fuse function", which ensures contiguity data is merged correctly. This now produces correct outputs:
image

### Closes issues (optional)
- Closes Issue #1337

### Checklist

sandbox_spellchecker

### DIFF
--- a/Tools/dea_tools/datahandling.py
+++ b/Tools/dea_tools/datahandling.py
@@ -17,7 +17,7 @@ here: https://gis.stackexchange.com/questions/tagged/open-data-cube).
 If you would like to report an issue with this script, you can file one
 on GitHub (https://github.com/GeoscienceAustralia/dea-notebooks/issues/new).
 
-Last modified: February 2025
+Last modified: April 2025
 """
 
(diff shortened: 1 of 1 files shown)


SpacemanPaul (approved):
Looks good to me.

PR #1346, opened 2025-05-22
Update USAGE.rst

cbur24:
Add recent citation from DE-Africa


GL-S:
I think currently the citations are in alphabetic order. I have moved the new citation below

GL-S (approved):

Issue #1115, opened 2023-08-29
Make `datacube` an optional dependency of `dea-tools`


robbibt:
Many functions in DEA Tools don't directly require `datacube` access (e.g. our coastal tide modelling functions, RGB and animation code, band indices, vectorising/rasterising tools etc). However, `datacube` is currently listed as a required dependency here:
https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Tools/setup.py#L37

To make `dea-tools` easier to install, we should consider making `datacube` an _optional_ dependency, using "extras" here:
https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Tools/setup.py#L68-L72

This should allow users to install `dea-tools` without having to also install `datacube`.

robbibt:
We should however ensure that some important dependencies like `odc-geo` and `odc-stac` are correctly included in our `setup.py`.

cbur24:
completed #1358

PR #1349, opened 2025-06-26
Use `uv` for managing Python package, simplify dependencies

robbibt:
### Proposed changes
This PR updates DEA Tools packaging with several major updates aimed at making DEA Tools easier to install both on and off the Sandbox.

- Use a more modern approach based on `uv` and a single `pyproject.toml` file
- Moved difficult-to-install Python packages like `datacube` into optional dependencies, so that users can install the package without needing access to a functional `datacube` environment like Sandbox
- Removed packages that aren't used by any notebook or script
- Add experimental config to support automated code formatting with `ruff`, and dependency checking with `deptry`

This is a work in progress, with more changes to be done once this PR is merged.

### Closes issues (optional)
- Closes Issue #1115


README_tools.md
```diff
@@ -0,0 +1,96 @@
...
+* `dea_tools.plotting`: Plotting tools for DEA data (e.g. RGB composites, animations, interactive maps)
+* `dea_tools.bandindices`: Functions to calculate remote sensing indices (e.g. NDVI, NDWI)
+* `dea_tools.spatial`: Spatial analysis utilities (e.g. rasterisation, vectorisation, contours, image processing)
+* `dea_tools.temporal`: Tools for temporal analysis (e.g. phenology, time-series statistics, multi-dimensional regression)
+* `dea_tools.classification`: Machine learning utilities (e.g. training and applying models on satellite data)
+* `dea_tools.dask`: Utilities for parallel processing with Dask (e.g. creating scalable Dask clusters)
+* `dea_tools.landcover`: Tools for accessing and visualising DEA Land Cover data
+* `dea_tools.coastal`: Coastal and intertidal analysis tools (e.g. tidal tagging, coastal change time series)
+* `dea_tools.bom`: Accessing Bureau of Meteorology water data (e.g. gauge and discharge data)
+* `dea_tools.climate`: Tools for working with gridded climate data (e.g. ERA5)
```
cbur24: This package is no longer in develop/dea-tools

Tools/dea_tools/temporal.py
```diff
@@ -434,8 +433,13 @@ def temporal_statistics(da, stats):
         temporal statistics
 
     """
+    # Attempt to import hdstats and raise an error if not available
+    try:
+        import hdstats
+    except ImportError:
+        raise ImportError("`hdstats` is required for `temporal_statistics`. Please install the `[hdstats]` dependency.")
```
cbur24, line 440: Its unfortunate that we need `hdstats` in this module when the (more commonly used) phenology and linear regression functions don't require it.

README_tools.md
```diff
@@ -0,0 +1,96 @@
...
+
+To work with this module on the DEA Sandbox or National Computational Infrastructure environments without installing it, you can add the `Tools` directory to the system path from within the `dea-notebooks` repository:
+
+```python
+import sys
+sys.path.insert(1, '../Tools/')
+import dea_tools.datahandling  # or some other submodule
+```
+
+You can also `pip install` the module directly from the local `Tools` directory. To do this on the DEA Sandbox, run `pip` from the terminal:
```
cbur24: From within the `/Tools` directory this didn't work for me: `no setup.py or pyproject.toml file found`.  Instead, needed to run `pip install -e .` from withing the `dea-notebooks` directory.

cbur24:
I was able to clone this branch of the repo and successfully run `pip install -e .`  Importing modules still fails though due to datacube dependecies (e.g. errant `from datacube.utils.cog import write_cog`, etc.) throughut the python scripts - are these to be resolved in the next PR?  A couple of the datacube dependecies are also solved through merging of #1351 

Otherwise happy to approve and see it progress through a bit of trail and error

Tools/dea_tools/temporal.py
caitlinadams: Which functions depend on hdstats? Would it make any sense to split functions that depend on hdstats into their own module? Or are all of them temporal/time-series functions that naturally belong together?

pyproject.toml
```diff
@@ -0,0 +1,230 @@
...
+
+[project.optional-dependencies]
+all = ["dea-tools[datacube,stac,jupyter,dask_gateway,hdstats,notebooks]"]
+datacube = [
+    "datacube>=1.8.5", 
+    "odc-ui>=0.2.1",
+    # "datacube-ows>=1.8.30",  # has unwanted psycopg2 dependency
+]
+stac = [
+    "odc-stac>=0.3.0",
```
caitlinadams:
```suggestion
    "odc-stac[botocore]>=0.3.0",
```

Suggest including botocore, since when using dea-tools, we might expect people to connect to DEA's AWS bucket. If botocore is not included, users will receive an error message saying they need to install it.

pyproject.toml
```diff
@@ -0,0 +1,230 @@
...
+    "pystac-client>=0.8.3",
+    "planetary_computer>=1.0.0",
+]
+jupyter = [
+    "jupyter>=1.0.0",
+    "ipyleaflet>=0.17.0",
```
caitlinadams, line 86: I wonder if it's worth adding ipykernel here to support running notebooks in VS Code? Users will receive a prompt from VS Code to install it if it's not present, so it's ok to not include it as a dependency (the prompt is more explicit than the botocore prompt I suggested above).

pyproject.toml
```diff
@@ -0,0 +1,230 @@
...
+    "intertidal",          # to fix with Sandbox upgrade
+    "osgeo",               # ignore due to confusion with gdal package
+]
+# Project should not contain unused dependencies
+DEP002 = [
+    "dea-tools",           # ignore dea-tools being mistaken for dependency
+    "jupyter",             # included for local environment creation
+    "pystac-client",       # included for local STAC support
+    "odc-stac",            # included for local STAC support
+    "planetary_computer",  # included for STAC support
```
caitlinadams, line 217: If you add botocore above, it might need to be added here too. This would also apply to ipykernel if you add it to the jupyter section.

caitlinadams (commented):
This is looking great -- thanks for putting in the effort to get this work started! I've made a few comments/suggestions, which you should feel free to include or not depending on your own thoughts.

pyproject.toml
robbibt: I think the `jupyter` metapackage should handle this - it installs the packages here:

https://github.com/jupyter/jupyter/blob/master/setup.py#L27

I'm still learning what specific parts of the Jupyter ecosystem we need to specify though - there's so many options!

pyproject.toml
robbibt: Great idea!

Tools/dea_tools/temporal.py
cbur24: Its only the [temporal_statistics](https://github.com/GeoscienceAustralia/dea-notebooks/blob/34d91d531909fb29800d25d0ecfbc29edb0b6c8d/Tools/dea_tools/temporal.py#L399) function that uses hdstats.  Conceptually it makes sense to be in the `temporal.py` file, but if it blocks usage of other, more widely used functions then I don't have a problem with splitting it out. Maybe this function should instead live within `odc-ago` where `hdstats` is already a dependency for the geomedian code-base?

caitlinadams:
Please forgive me if I missed this when having a skim through the codebase, but do you have an automated method for generating the conda-environment.yaml file? 

When working with `pixi` in our sar-pipeline work, I used `pixi`'s [tasks functionality](https://pixi.sh/dev/workspace/advanced_tasks/) to define a task that [exports the conda environment](https://github.com/GeoscienceAustralia/sar-pipeline/blob/7fae2fb7d2403ce3f3966040cf0c6972d8e427eb/pyproject.toml#L106-L109). I then wrote a [pre-commit hook](https://github.com/GeoscienceAustralia/sar-pipeline/blob/main/hooks/export_conda.sh) that will [execute that task](https://github.com/GeoscienceAustralia/sar-pipeline/blob/7fae2fb7d2403ce3f3966040cf0c6972d8e427eb/.pre-commit-config.yaml#L15-L20), which means the `environment.yaml` we provide is always in sync with the `pyproject.toml`. 

I was mostly curious if uv has a similar task functionality, and how you're managing the conda environment file.

pyproject.toml
caitlinadams: Perfect :)

robbibt:
> Please forgive me if I missed this when having a skim through the codebase, but do you have an automated method for generating the conda-environment.yaml file? 
> 
> When working with `pixi` in our sar-pipeline work, I used `pixi`'s [tasks functionality](https://pixi.sh/dev/workspace/advanced_tasks/) to define a task that [exports the conda environment](https://github.com/GeoscienceAustralia/sar-pipeline/blob/7fae2fb7d2403ce3f3966040cf0c6972d8e427eb/pyproject.toml#L106-L109). I then wrote a [pre-commit hook](https://github.com/GeoscienceAustralia/sar-pipeline/blob/main/hooks/export_conda.sh) that will [execute that task](https://github.com/GeoscienceAustralia/sar-pipeline/blob/7fae2fb7d2403ce3f3966040cf0c6972d8e427eb/.pre-commit-config.yaml#L15-L20), which means the `environment.yaml` we provide is always in sync with the `pyproject.toml`. 
> 
> I was mostly curious if uv has a similar task functionality, and how you're managing the conda environment file.

The Conda environment mentioned in the readme actually comes from `datacube-core`. I don't know if it works, but I left it in there in case it's useful. It used to be the only way to create an environment with functioning GDAL.

At the moment we don't do anything Conda relatated in `dea-tools` - it's all purely `pip`. It definitely could be better to provide a Conda env specific to `dea-tools` instead of using the ODC one - it's mainly that I have no Conda experience 🙂

caitlinadams:
> The Conda environment mentioned in the readme actually comes from `datacube-core`. I don't know if it works, but I left it in there in case it's useful. It used to be the only way to create an environment with functioning GDAL.
> 
> At the moment we don't do anything Conda relatated in `dea-tools` - it's all purely `pip`. It definitely could be better to provide a Conda env specific to `dea-tools` instead of using the ODC one - it's mainly that I have no Conda experience 🙂

That makes sense -- I hadn't put that together (it explains why the default env it creates is called cubeenv!), but I can now see that the URL it's collected from is datacube-core, rather than this repo. 

Reviewing that conda environment, it doesn't look sufficient to support dea-tools, but it should be checked. For example, it doesn't contain packages like `fiona` and `seaborn`, and contains a lot of extra stuff that dea-tools users won't need (like `sphinx`). 

As a stop-gap in the `README_tools.md` file, I would recommend adding an extra line after activating the conda environment to pip install dea-tools. That way, the user will get what they need. I think there could be a broader discussion about whether you want to include conda support at all. Pixi supports both conda and pip, which is why we've been using it for `sar-pipeline` (we need gdal and didn't want to have to go through the process of installing it seperate to our package).

README_tools.md
```diff
@@ -0,0 +1,96 @@
...
+
+### With conda
+
+If you encounter issues with the installation, try installing the package in a `conda` Python environment where [GDAL](https://pypi.org/project/GDAL/) and [pyproj](https://pypi.org/project/pyproj/) are already installed:
+
+```console
+wget -O conda-environment.yml https://raw.githubusercontent.com/opendatacube/datacube-core/develop/conda-environment.yml
+
+mamba env create -f conda-environment.yml
+conda activate cubeenv
```
caitlinadams, line 68:
```suggestion
conda activate cubeenv
pip install dea-tools
```

This will ensure that any packages that aren't installed from the datacube core conda environment are added via pip. But we should definitely test this before adding it. Let me know if you want me to test it for you -- I'm primarily a Pixi/conda user these days.

robbibt: If you had time to try it out, that would be really valuable - I've always been a bit unsure of whether any of these Conda instructions work, but haven't been able to test it myself.

pyproject.toml
robbibt: `uv run deptry .` seems to run without any issues, so seems OK for now

Tools/dea_tools/temporal.py
robbibt: Because the module import is now inside the `temporal_statistics` function, I don't think it should block importing other functions from the module any longer - the error will only be returned if specifically `temporal_statistics` is imported without `hdstats` installed.

This is an approach I want to do for `datacube` too - move any of those optional imports to the function level, vs. module level.

README_tools.md
jennaguffogg: I've been navigating to the Tools directory and using -e . - so same tactic as Chad but for me it works in the Tools directory.

README_tools.md
```diff
@@ -0,0 +1,97 @@
...
+```python
+from dea_tools.datahandling import load_ard
+from dea_tools.plotting import rgb
+```
+
+## Citing DEA Tools
+
+If you use any of the notebooks, code or tools in this repository in your work, please reference them using the following citation:
+
+> Krause, C., Dunn, B., Bishop-Taylor, R., Adams, C., Burton, C., Alger, M., Chua, S., Phillips, C., Newey, V., Kouzoubov, K., Leith, A., Ayers, D., Hicks, A., DEA Notebooks contributors 2021. Digital Earth Australia notebooks and tools repository. Geoscience Australia, Canberra. https://doi.org/10.26186/145234
```
jennaguffogg: This is a 2021 citation, is there a more updated version? Do we do one of these each calendar year?

README_tools.md
jennaguffogg: Same here. I tried using `pip install -e .` both inside the noteboooks directory and also navigating into the `/Tools` directory. `pip install -e .` inside `/Tools` didn't work, but it did work in the parent directory like it did for Chad.

README_tools.md
```diff
@@ -0,0 +1,97 @@
...
+
+To install with other packages used in DEA Notebooks examples:
+```console
+pip install dea-tools[jupyter,dask_gateway,hdstats,notebooks]
+```
+
+Or to install with *all* optional dependencies:
+```console
+pip install dea-tools[all]
+```
```
jennaguffogg, line 58: I tested this in vs code with a virtual environment. Hit an error with psycopg2:

Collecting psycopg2 (from datacube-ows->dea-tools)
  Using cached psycopg2-2.9.10.tar.gz (385 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [34 lines of output]
      /tmp/pip-build-env-022o0uo9/overlay/lib/python3.13/site-packages/setuptools/dist.py:759: SetuptoolsDeprecationWarning: License classifiers are deprecated.
      !!
      
              ********************************************************************************
              Please consider removing the following classifiers in favor of a SPDX license expression:
      
              License :: OSI Approved :: GNU Library or Lesser General Public License (LGPL)
      
              See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
              ********************************************************************************
      
      !!
        self._finalize_license_expression()
      running egg_info
      writing psycopg2.egg-info/PKG-INFO
      writing dependency_links to psycopg2.egg-info/dependency_links.txt
      writing top-level names to psycopg2.egg-info/top_level.txt
      
      Error: pg_config executable not found.
      
      pg_config is required to build psycopg2 from source.  Please add the directory
      containing pg_config to the $PATH or specify the full executable path with the
      option:
      
          python setup.py build_ext --pg-config /path/to/pg_config build ...
      
      or with the pg_config option in 'setup.cfg'.
      
      If you prefer to avoid building psycopg2 from source, please install the PyPI
      'psycopg2-binary' package instead.
      
      For further information please check the 'doc/src/install.rst' file (also at
      <https://www.psycopg.org/docs/install.html>).
      
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
(.packagetesting) dea-notebooksubuntu@:~/repos/dea-notebooks$

Tools/dea_tools/temporal.py
cbur24: Ah cool, nice and simple solution - I like it.

README_tools.md
robbibt: Just this one - it ties back to an `ecat` listing so it's locked in. However, I think we should think about publishing an official Journal of Open Source Software paper for DEA Notebooks!

README_tools.md
robbibt: Yeah, this is unfortunately probably expected at this point. Issues like this are why those dependencies were moved to optional: packages like `psycopg2` (which is a `datacube` dependency) are really difficult to install. This PR doesn't really fix those core installability issues, but lets you bypass them more easily.

For this specific issue, I raised an issue here: https://github.com/opendatacube/datacube-ows/issues/1178 (@SpacemanPaul would making  `psycopg2` an optional dependency in `datacube-ows` be a tricky task? It would definitely make it easier to install DEA Tools...)

For now @jennaguffogg, the best solution if you can't install `psycopg2` would be to install DEA Tools without the `datacube` dependency, e.g. `dea-tools[stac,jupyter,dask_gateway,hdstats,notebooks]` instead of `dea-tools[all]`

robbibt: I think I'll add a note here to say that `dea-tools[all]` might not work for everyone.

Tests/test_notebooks.sh
jennaguffogg, line 1: on line 81, is `fi` a typo?

robbibt: I also thought that looked weird, but apparently that's the correct way to close an if/else statement in bash!
https://bash.cyberciti.biz/guide/If..else..fi

cbur24 (approved):
Great work, Robbi.

PR #1350, opened 2025-06-26
Fix issue with loading climate data over THREDDS

cbur24:
### Proposed changes
For unknown reasons, loading data over NCI THREDDS using dask has begun failing with an I/O error. This change moves the dask client to below the ANUCLIM section to resolve the issue.


robbibt (approved):
Looks great, fingers crossed this fixes it!

PR #1352, opened 2025-07-01
Add WIT app to README.rst

LaurenSchenk1:
Update rst to include WIT app so it will appear in interactive apps.


margaretharrison (approved):
Looks good to me!

PR #1360, opened 2025-07-15
Update `create_local_dask_cluster` to work without `datacube` dependency

cbur24:
### Proposed changes
As per the title.  Ran a couple of notebooks that use this function to confirm its working.  The defaults will provide almost the same experience as the previous function (1 worker, with number of threads equal to the number of cpus), only there is no longer the option of providing the `spare_mem` parameter. Which is not used within `dea-notebooks` anyway. This does mean the cluster gets access to all the memory on the machine.


Tools/dea_tools/dask.py
```diff
@@ -41,44 +46,46 @@ def create_local_dask_cluster(spare_mem="3Gb", display_client=True, return_clien
         sys.path.append("../Scripts")
```
robbibt: "Scripts" no longer exists, can we change this to just include the import and not the `sys` bit?

Tools/dea_tools/dask.py
```diff
@@ -41,44 +46,46 @@ def create_local_dask_cluster(spare_mem="3Gb", display_client=True, return_clien
...
         An optional boolean indicating whether to display a summary of
         the dask client, including a link to monitor progress of the
         analysis. Set to False to hide this display.
     return_client : Bool, optional
         An optional boolean indicating whether to return the dask client
         object.
+    n_workers : int, optional
+        Number of workers to start, default is set to 1 which works well
+        with loading ODC data.
+    threads_per_worker: int, optional
```
robbibt, line 68: This is great - can we also add a `**kwargs` here that gets passed to the `dask.distributed.Client` call at line 82? That way users can do even more custom stuff if necessary

Tools/dea_tools/dask.py
```diff
@@ -41,44 +46,46 @@ def create_local_dask_cluster(spare_mem="3Gb", display_client=True, return_clien
...
+        else:
+            threads_per_worker = os.cpu_count()
+
+    # start client
+    client = dask.distributed.Client(
+        n_workers=n_workers, threads_per_worker=threads_per_worker
+    )
+    
+    # configure aws access
     if _IS_AWS:
```
robbibt: I wonder if this is broadly applicable outside of the Sandbox (e.g. if a user is running code locally or in Github Codespaces etc - I might ask whether there's any reason for this check

robbibt (changes requested):

Tools/dea_tools/dask.py
cbur24: I'm not really sure about this. Only included because it was there in the old function so I assumed it must be checking some environment variable.

Tools/dea_tools/dask.py
cbur24: added kwargs (hopefully correctly)

robbibt (approved):
Looks great - this will be a really nice addition to v0.4.0.

I made a change to add `configure_rio` as a configurable param. I think this should be pretty safe, but at least users can easily turn it off as need be.

Will wait to tests pass before merging.

cbur24:
Thanks for the prompt and thorough review @robbibt!

emmaai (commented):
It might not a good idea to remove default `memory_limit`, the old function buffers the memory a little, reduces the chance of "oom".

cbur24:
Thanks for the review @emmaai. Are you able to advise on how to add back the `spare-mem` param without necessarily reintroducing all of this code? https://opendatacube.readthedocs.io/en/latest/_modules/datacube/utils/dask.html#start_local_dask

robbibt:
This is what gets called by default: `memory_limit="auto"`: https://distributed.dask.org/en/stable/api.html#distributed.LocalCluster
image

@emmaai Do you have a simple rule of thumb/code you use to configure memory? Maybe we could borrow that here.

I guess we could just copy the old functions into this module completely; feels a bit bad to duplicate code, but we could add a clear note that it's also available in `datacube-core`. Would be nice to keep this simple though.

cbur24:
Since the default is to have only 1 worker, then perhaps a float value like `memory_limit=0.95` would work well? That way we always have 5% left over for jupyterlab etc?

emmaai:
0.9 or 0.95/n_workers could be a good guess, `auto` will deem all the memories excluding os as available.

cbur24:
Default is to now split 95% of system memory by n_workers.
```
memory_limit: str, float, int, or None, optional
        Sets the memory limit per worker. Default is 'spare_mem', where 95 % of the available
        system memory is split among the number of workers, allowing spare memory to be withheld
        from the cluster.
        
```

PR #1363, opened 2025-07-23
Tools for random sampling of xarray dataarrays

cbur24:
### Proposed changes
This PR introduces a Python function and an accompanying Jupyter notebook for efficiently sampling pixels from xarray.DataArray objects using various sampling strategies, including:
- Random sampling
- Stratified random sampling
- Equal stratified random sampling
- Manual specification of samples per class

These tools are designed to support the extraction of training and validation datasets.

I've also made a small change to the notebooks template to remove reference to the 'Scripts/' folder

### Checklist


robbibt:
Hey @cbur24, this looks brilliant! I'll have a look hopefully today or tomorrow. A few quick thoughts:

- We probably need to add the new notebook to the readme index file for Knowledge Hub
- Would it be worth investigating about adding some simple unit tests for the function? E.g. checking that `n` points are successfully returned, that the ratios between the classes are correct etc for the various sampling methods. There's [some examples here](https://github.com/GeoscienceAustralia/dea-notebooks/blob/sampling-tools/Tests/dea_tools/test_temporal.py) that could be used as a starting point. Just thinking that this function is probably something we'll want to use operationally, and it would be great to have some assurance everything works as expected.

cbur24:
> Hey @cbur24, this looks brilliant! I'll have a look hopefully today or tomorrow. A few quick thoughts:
> 
> * We probably need to add the new notebook to the readme index file for Knowledge Hub
> * Would it be worth investigating about adding some simple unit tests for the function? E.g. checking that `n` points are successfully returned, that the ratios between the classes are correct etc for the various sampling methods. There's [some examples here](https://github.com/GeoscienceAustralia/dea-notebooks/blob/sampling-tools/Tests/dea_tools/test_temporal.py) that could be used as a starting point. Just thinking that this function is probably something we'll want to use operationally, and it would be great to have some assurance everything works as expected.

Sounds good @robbibt, thanks for reviewing.
- Can you point me to where I need to edit KH?
- Unit tests for this is a good idea. I'll look into it over this week

robbibt:
Just here: https://github.com/GeoscienceAustralia/dea-notebooks/blob/sampling-tools/How_to_guides/README.rst

robbibt (approved):
Looks great!

jennaguffogg (approved):
I tested the markdown code and it ran great!

PR #1362, opened 2025-07-16
Add new submodule for continental COG and VRT generation

GL-S:
### Proposed changes
Code to generate continental COGs and VRTs to add colour scheme (or RGB) was added.
As well as json files with colour scheme dictionaries for land cover.
The code is generic enough that can be used for any summary product. It is used by Argo workflows

PR #1364, opened 2025-07-24
Updates to continental mosaics PR

robbibt:
### Proposed changes
This PR makes a number of updates to the COG mosaicking code in #1362:

- Removed custom install of DEA Tools (once the Sandbox image is updated it will be installed by default)
- Renamed `compression_lvl` to `compression_level` for readability
- Used CLI utility commands by default in readme, instread of `python ...` approach
- Moved checks for S3 up to the top of functions, and re-used `_is_s3` function rather than copying it into both scripts
- Only use relative VRT links if COGs are local. Use absolute links if COGs are on S3.
- Added new `vrt_name` param to allow passing in a custom name for VRTs instead of using the default `{red-green-blue}` band names (e.g. `ga_s2_tidal_composites_cyear_3_2022_vrt-low-truecolour.vrt`)
- Fixed bug where `list_tiles` was not actually filtering the tile list
- Use `configure_rio` from `odc.stac` instead of `configure_s3_access` from datacube
- Formatted and linted code with Ruff, minor docs updates

### DIFF
--- a/How_to_guides/Generating_COG_mosaics.ipynb
+++ b/How_to_guides/Generating_COG_mosaics.ipynb
@@ -20,7 +20,7 @@
    "source": [
     "## Background\n",
     "\n",
-    "DEA products are provided as tiles. The `datacube` library allows users to seamlessly work with these tiles and automatically mosaic them together for the area and time of interest. The data array loaded can then be exported as a standalone file.\n",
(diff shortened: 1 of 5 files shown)


colourmeamused:
Hi Robbi! This line has a typo, might be better to fix here then merge into [continental_mosaics](https://github.com/GeoscienceAustralia/dea-notebooks/tree/continental_mosaics)

https://github.com/GeoscienceAustralia/dea-notebooks/blob/5b0b5b2feb44b77e22f366d300f51efa3d5f2f50/Tests/test_notebooks.sh#L42

GL-S (approved):
great changes, thanks Robbi

PR #1368, opened 2025-07-30
Modified continental COG mosaics submodule to contain colour schemes for VRTs

GL-S:
Previously, the colour schemes were in `Supplementary materials`. It worked fine within the sandbox, but not if dea-tools was installed and used outside of it.

Now the colour schemes are contained and extracted from a python script (inside the module) with all styling dictionaries, rather than individual json files.

It is still possible to use a custom colour scheme by providing the path to a json file structured similarly to a python dictionary. I have mentioned this functionality in the example notebook and expanded the usage explanation in the functions and CLI descriptions. I have not included a practical demonstration in the notebook for sake of brevity, but I have tested it and worked well.


robbibt (approved):

cbur24 (approved):
LGTM!

PR #1370, opened 2025-08-08
adapted-LC-to-use-STAC

GL-S:
### Proposed changes
I have updated all land cover notebooks and relevant code to use STAC rather than datacube's dc.load().

The changes made to code in landcover.py are only for the plotting of land cover. 
I have tested the DEA_Land_Cover.ipynb notebook that uses dc.load() with these recent changes, and the workflow is not broken. 
So both the STAC and the datacube approach are still possible.

I have also made some minor changes to formatting and spelling.

### Checklist

sandbox_spellchecker


GL-S:
For some reason the figure in the pixel drill is different for the two methods for the same year, while all other plots and animations in the other notebooks are identical. I will need to investigate further to understand what's going on

GL-S:
> For some reason the figure in the pixel drill is different for the two methods for the same year, while all other plots and animations in the other notebooks are identical. I will need to investigate further to understand what's going on

Does odc.stac assign the central date to annual products, or the first day of the year?
datacube uses the central date, so LC for year 2020 is assigned to the 1st or 2nd July 2020.

I think by filtering using the year only and nearest as method (as we do in the pixel drill notebook) I am actually calling different years depending on if I use datacube or odc.stat.


**EDIT**: yes, that's what's happening. With odc.stac is actually more correct! 
I think with datacube we were selecting 2019 instead.

GL-S:
Also, should we keep the STAC versions of the notebook separate, like in a subfolder? Or is it ok to replace the datacube ones?

I am also thinking that maybe I shouldn't have modified the Microsoft Planetary Computer one, as it was also a comparison of methos on how to access different data for the same region and time period.

cbur24:
@GL-S @robbibt @caitlinadams 

Maybe we should have some standardised text at the top of each notebook that loads with odc-stac? Something like:

⚠️ **Important Note**
This notebook loads data into xarray using [odc-stac](https://odc-stac.readthedocs.io/en/latest/intro.html). This means:
1. The notebook can be run either within the sandbox, or on any computer with an internet connection and the required Python packages installed (e.g. by running **pip install dea-tools** in your Python environment).
2. Data loading performance will depend on the available internet bandwidth.

cbur24:
> Also, should we keep the STAC versions of the notebook separate, like in a subfolder? Or is it ok to replace the datacube ones?
> 
> I am also thinking that maybe I shouldn't have modified the Microsoft Planetary Computer one, as it was also a comparison of methos on how to access different data for the same region and time period.

No, I believe the idea is to eventually replace all notebooks with the odc-stac versions, thereby freeing dea-notebooks from the sandbox

GL-S:
> @GL-S @robbibt @caitlinadams
> 
> Maybe we should have some standardised text at the top of each notebook that loads with odc-stac? Something like:
> 
> ⚠️ **Important Note** This notebook loads data into xarray using [odc-stac](https://odc-stac.readthedocs.io/en/latest/intro.html). This means:
> 
> 1. The notebook can be run either within the sandbox, or on any computer with an internet connection and the required Python packages installed (e.g. by running **pip install dea-tools** in your Python environment).
> 2. Data loading performance will depend on the available internet bandwidth.

Yes, some kind of warning is a good is idea.

robbibt:
Yeah, I think some standards and templating will be really important. It might even be a good opportunity to refresh the entire top block of the Notebooks! Especially now we have things like Knowledge Hub, and if in the future we want to point out to options like Google Colab/Binder etc.

Just a general question: is there a short deadline for getting these changes merged? It just might be good to have a DEA Notebooks catchup soon to discuss how they'll fit with the broader STAC refresh plans, and see if we can make them as standardised and streamlined as possible.

cbur24:
> Yeah, I think some standards and templating will be really important. It might even be a good opportunity to refresh the entire top block of the Notebooks! Especially now we have things like Knowledge Hub, and if in the future we want to point out to options like Google Colab/Binder etc.
> 
> Just a general question: is there a short deadline for getting these changes merged? It just might be good to have a DEA Notebooks catchup at some point to discuss how they'll fit with the broader STAC refresh plans, and see if we can make them as standardised and streamlined as possible.

No pressing deadlines on this, was just something that came up so we progressed it.  I agree - let's set a task for updating the DEA-Notebooks template with new info on stac, links to binder etc.

robbibt:
> > Also, should we keep the STAC versions of the notebook separate, like in a subfolder? Or is it ok to replace the datacube ones?
> > 
> > I am also thinking that maybe I shouldn't have modified the Microsoft Planetary Computer one, as it was also a comparison of methos on how to access different data for the same region and time period.
> 
> No, I believe the idea is to eventually replace all notebooks with the odc-stac versions, thereby freeing dea-notebooks from the sandbox

There's pros and cons of both approaches - personally I think having duplicates of every notebook will get unwieldy very fast, and make maintenance really difficult at the scale of the entire repo (it's already a lot of work even with our current notebooks). So I'd probably lean towards not duplicating, but nothing is set in stone yet.

robbibt:
> No pressing deadlines on this, was just something that came up so we progressed it.  I agree - let's set a task for updating the DEA-Notebooks template with new info on stac, links to binder etc.

It's super neat, and having some working examples like this will make it so much easier to start doing this for the full repo - very exciting!

PR #1372, opened 2025-08-11
Develop to Stable

GL-S:
The main reason I am creating this PR is because the default notebooks and code in the Sandbox are becoming obsolete. Especially the ones related to land cover use old functions and logic. 

The LC notebooks can still run, however plotting any area with no-data pixels will result in wrong colours (this is because the no-data value changed between LC versions, and the current code in stable uses the old colour scheme).
It would also be convenient for some quick tests without needing to clone develop.


cbur24 (approved):

robbibt (approved):
Looks great @GL-S! When you're ready to merge, can we please do it with this option - it makes it easier to keep develop and stable in sync:

Image

PR #1387, opened 2025-09-03
Additional notebook template updates

robbibt:
A few suggested changes to the template:

- Add additional optional boxes in the header
- Combine STAC catalogue and S3 access sections
- Remove tags - they aren't used in Knowedge Hub any more as far as I know, do we still need them?


cbur24 (approved):

cbur24:
Nice we don't need to mantain tags anymore

PR #1390, opened 2025-09-05
geobox-fixes-LC

GL-S:
quick fix around use of geobox. This new method is robust and works with both STAC and datacube
All Landcover notebooks were re-run and nothing broke

the look of those big dictionaries at the top of landcover.py has also been tidied up


cbur24 (approved):
Thanks @GL-S

GL-S:
this is failing some tests due to the geomad module

Tools/dea_tools/landcover.py
```diff
@@ -19,7 +19,7 @@
 
 Last modified: May 2025
 """
-
+import odc.geo
```
robbibt: I think you may need to import `odc.geo.xr` to get the `.odc` accessor added to your xarray (not just `odc.geo`)

GL-S: I see, but the code works though. Is this related to the failed tests?

robbibt: It's more that this line is doing nothing in its current state (I can't see it used anywhere in the module). 

Datacube will also automatically add `.odc` to any loaded data, but it can be good to import `odc.geo.xr` to just double check/ensure it's added if it's missing for whatever reason.

GL-S: oh I see, ok I will edit this. Together with pull the latest changes from the last PR.
Strange that it was working anyway. I thought that was for making `data.odc.geobox.shape` work

robbibt: You can see more info here: https://odc-geo.readthedocs.io/en/latest/intro-xr.html

> Importing odc.geo.xr enables .odc extension on [xarray.DataArray](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.html#xarray.DataArray) and [xarray.Dataset](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.html#xarray.Dataset).

I suspect it was working because `datacube` and `odc-stac` effectively do the same thing under the hood.

GL-S: yes I think you are right, I tested the code without that import at all, and it works anyway. So it doesn't even seem to be needed, but I changed it to odc.geo.xr and kept it just to be sure

PR #1393, opened 2025-09-11
Add `NIRv` vegetation index and tidy up of notebook

cbur24:
### Proposed changes
Adding an important vegetation index to `band_indices.py`, along with tidying up the notebook, and also added a `verbose=True` parameter so the user can opt to remove the 'dropping bands...' print statements.


robbibt (approved):
Looks great - just re-ran the failing tests in case it was something ephemeral. Ready to merge if they pass this time!

PR #1402, opened 2025-09-25
Fractional cover notebook upgrade to odc 1.9

jennaguffogg:
This PR includes a warnings wrapper for `dc.load_product()`, though this currently has no impact. When odc-geo GridSpec warnings are moved to class deprecation, this will suppress them.

loading of wofs product modified to include time parameter, as odc 1.9 changes behaviour of the `dc.load(like=...)` to only apply to the spatial extent/pixels of the provided dataset.


robbibt:
Hey @jennaguffogg, I think the test failure here means you need to restart the notebook and "Run all cells" so all the cells have been run in sequence and have outputs. 🙂

DEA_products/DEA_Fractional_Cover.ipynb
```diff
@@ -87,538 +87,6 @@
    "cell_type": "code",
```
robbibt, line 87: <pre class="ql-syntax" spellcheck="false">Line #1.    # List DEA Fractional Cover products available in DEA
</pre><p>The "with warnings.catch_warnings():" code doesn't seem to have actually stopped the warnings, so we might want to get rid of it to keep things simple. Hopefully there will be a better fix soon</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/GeoscienceAustralia/dea-notebooks/pull/1402/discussion/#comment-2378192373'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

GL-S, line 87: <p>I think the ";" could be re-introduced here, to avoid printing "Text(0, 0.5, 'Percent cover (%)')"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/GeoscienceAustralia/dea-notebooks/pull/1402/discussion/#comment-2380556670'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

robbibt (approved):
For now I removed the warning supression and instead re-ran the notebook on "default" branch on the Sandbox to remove the warnings - we can revisit this once Ariana's fix is implemented, and perhaps apply it to all the DEA product notebooks consistently. For now though I think we can merge this as-as - thanks Jenna!

PR #1409, opened 2025-10-02
Phenology and temporal statistics notebook

cbur24:
### Proposed changes

The `dea-tools.temporal.temporal_statistics` function fails on the unstable sandbox image because the version of `hdstats` we use is compiled against numpy v1, while the unstable sandbox is now using numpy v2.  
This PR is to first discuss, and then implement whatever option we think is best. Our options include:

1. Adapting the functions in `hdstats.ts.pyx` to xarray and adding them to dea-tools,
2. Drop this functionality from dea-notebooks altogether. 
3. Install and compile `hdstats` directly from the repo: https://github.com/daleroberts/hdstats rather than using our private version (possible now that GM has been moved to its own public repo).

EDIT: Testing adding the relevant stats to `dea-tools.temporal`, thereby removing any dependency on `hdstats`. Note that installing `hdstats` directly (option 3) doesn't work as the repo's installation is broken and has been for several years.


robbibt:
Hey @cbur24, this looks fantastic. Would it be possible to get a short docstring for the new funcs describing what they do? Even just the snippets from the main `temporal_statistics` docstring would be great.

robbibt (approved):
This looks fantastic, will be great to simplify our dependencies. Can you also remove these lines here? 

https://github.com/GeoscienceAustralia/dea-notebooks/blob/test_veg_pheno/pyproject.toml#L94-L96

There might be a line in our DEA Tools readme that mentions hdstats too.

Edit: tests!! So good 🤩🚀

robbibt:
This might also need updating to remove `hdstats`: https://github.com/GeoscienceAustralia/dea-notebooks/blob/test_veg_pheno/pyproject.toml#L79

PR #1411, opened 2025-10-10
Add Intro to Fuel Moisture Content notebook to DEA Notebooks

biancafichera:
### Proposed changes
The DEA Fuel Moisture Content product and Knowledge Hub page is currently live, so it's time for us to also officially upload the Introduction to DEA Fuel Moisture Content notebook. 
This notebook was originally reviewed by many in a draft pull request, but we created a new branch for the file name convention changes so this is a new pull request and comes from a different branch. All suggestions in the draft pull request have been changed here and this is the most up to date file.

### Checklist

sandbox_spellchecker


cbur24:
Looks great @biancafichera! I made a few minor edits to the markdown text, and a few comments below that you could consider addressing.
 
- Recommend using a DEA Maps link that has FMC pre-loaded.
- When run on the unstable sandbox (ODC-1.9) the `list_products` cell produces many ugly warnings `In odc-geo GridSpec, resolution is expected in (X, Y) order, or simply X if using square pixels with inverted Y axis.` Not sure if you can resolve this as I think its a broader issue but I thought it was worth noting. Maybe you should remove the `list_products` and just keep `list_measurements` since there is only one product anyway?
- Consider reducing the size of the area so it loads a little quicker - kinda slow with loading an entrire year of S2.  I tried adding a progress bar to `dc.load` but it doesnt seem to work on odc-1.9.

cbur24 (approved):

GL-S:
About the last figure, as Chad mentioned it is not a bug, but an expected behaviour. 
If you prefer a more consistent look of the x axis, you could do something like this:

```
fig, ax = plt.subplots(figsize=(12, 4))

monthly_mean.plot(ax=ax)
ax.set(
    xlabel=None,
    ylabel="FMC (%)",
    title="Average monthly FMC in \n Wiarborough Nature Reserve and Surrounds for 2021",
)

x_ticks = pd.to_datetime(monthly_mean.time.values)
ax.set_xticks(x_ticks)
ax.set_xticklabels(x_ticks.strftime("%b %Y"), rotation=40)

ax.grid(alpha=0.5);
```


And it would look like the following:
image

PR #1426, opened 2025-11-06
Coreg

arcisad:
### Proposed changes
This is my first attempt to create a notebook here, so I imagine I will be needing a lot of help and feedback :)

This is a notebook for running co-registration tools on satellite scenes. This particular one, co-registers two Sentinel-2 scenes against a reference Landsat-8 one.

I used AROSICS, KARIOS and Co-register tools to carry out the process. The first two are publicly available packages and the third one is the one I developed. 

*Note:* I have developed my own utility functions in DEA_tools/coreg.py file for the co-registration process as the initial plan was not to integrate with ODC libraries.  I am sure a lot of similar and even better functions are already available in ODC libraries but the ones I created here are specifically tailored to the requirements of my co-registration process. I appreciate your feedback on these utilities, whether it is about the structure of the code or any better already available alternative.  :)  

### Checklist

- [ x] Checked the structure of the notebook follows our [DEA-notebooks template](https://github.com/GeoscienceAustralia/dea-notebooks/wiki/NotebookTemplate)
- [ x] Removed any unused Python packages from `Load packages`
- [x ] Removed any unused/empty code cells
- [x ] Removed any guidance cells (e.g. `General advice`)
- [x ] Ensured that all code cells follow the [PEP8 standard](https://www.python.org/dev/peps/pep-0008/) for code. The `jupyterlab_code_formatter` tool can be used to format code cells to a consistent style: select each code cell, then click `Edit` and then one of the `Apply X Formatter` options (`YAPF` or `Black` are recommended).
- [ x] Included relevant tags in the final notebook cell (refer to the [DEA Tags Index](https://knowledge.dea.ga.gov.au/genindex/), and re-use tags if possible)
- [ x] Tested notebook on the [DEA Sandbox](https://knowledge.dea.ga.gov.au/guides/setup/Sandbox/sandbox/)
- [ x] Cleared all outputs, run notebook from start to finish, and save the notebook in the state where all cells have been sequentially evaluated
- [x ] If applicable, update the `Notebook currently compatible with` line below the notebook title to reflect the environments the notebook is compatible with
- [ x] Check for any spelling mistakes using the DEA Sandbox's built-in spellchecker (double click on markdown cells then right-click on pink highlighted words). For example:


abradley60:
Hey @arcisad thanks for this, I am starting to test some of the coregistration functions with some data I am playing with. I'm not the right person to guide how this should look for the notebook but I'm wondering if we should some spend time to enable coregistration through xarray, and make it available as part of the `dea-tools` package. For example, how the [xr_optical_flow](https://knowledge.dea.ga.gov.au/notebooks/Real_world_examples/Tracking_sediment_flow/#Background) function and (code)[https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Tools/dea_tools/temporal.py#L1257] works. 

Whether this is worth the time I am unsure, but it would be handy to have something like `xr_coreg` in the tests I'm doing now.

arcisad:
> Hey @arcisad thanks for this, I am starting to test some of the coregistration functions with some data I am playing with. I'm not the right person to guide how this should look for the notebook but I'm wondering if we should some spend time to enable coregistration through xarray, and make it available as part of the `dea-tools` package. For example, how the [xr_optical_flow](https://knowledge.dea.ga.gov.au/notebooks/Real_world_examples/Tracking_sediment_flow/#Background) function and (code)[https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Tools/dea_tools/temporal.py#L1257] works.
> 
> Whether this is worth the time I am unsure, but it would be handy to have something like `xr_coreg` in the tests I'm doing now.

Eventually yes, I think this should be the goal. But I am not sure this is the priority right now. Because I don't know if and when we want to turn co-reg into a user ready package. I have an outstanding ticket to discuss this for some time now.  :)

PR #1443, opened 2025-12-10
Update DEA Coastlines notebook to use new template

robbibt:
- [x] Notebook updated to follow [updated DEA Notebooks template format](https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop_stac/DEA_notebooks_template.ipynb)
- [ ] For STAC loading:
    - [ ] odc-stac `load()` and `configure_s3_access` imported
    - [ ] `pystac_client` `Client` imported
    - [ ] Environment setup includes `catalog`, `stac_client` and `configure_s3_access()`
    - [ ] Cell that searched a collection for items with `pystac-client`
    - [ ] Cell that loads discovered items into an xarray with `odc-stac`
- [x] Notebook no longer requires `datacube`
- [x] Notebook no longer does `sys.path.insert(1, '../Tools/')`


margaretharrison (approved):
Looks great! Had to pip install mapclassify to work in codespace.

PR #1468, opened 2026-02-01
Workflows/configure linters

GA-XavierGonzalez:
### Proposed changes
Removed push and pull request invocation from Ruff linter. This means the linter will only run manually, we can then have a look at the output and configure some usable rules as required.

Also hotfixed Bandit so it correctly scans ipynb files and uploads findings to the security tab.


cbur24 (approved):

stephenohairga:
LGTM, thanks Xavier.

robbibt:
Hey @GA-XavierGonzalez! Just back online today so a little late to the party. Just wanted to check - will this change integrate with our existing Ruff configuration in our `pyproject.toml` here? https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/pyproject.toml#L153-L187

PR #1484, opened 2026-04-17
collect_data_from_shp_more_robust

GL-S:
### Proposed changes
Small change to make collect_training_data more robust. 

Previously, if the feat_func or the query passed to collect_training_data included setting the output CRS to EPSG:4326 for example, the function _get_training_data_for_shp would fail when searching for x and y.

Now it first checks if the ds coords are x/y. If not, it extracts latitude/longitude

Also, I added a note to the doc string of collect_training_data to remind that, if the query includes a lat/lon output_crs, also the resolution needs to be in degrees. As this gave me some trouble when testing the changes made and getting unexpected results, so I thought it could be useful for users too. 

### Checklist

- [na] Checked the structure of the notebook follows our [DEA-notebooks template](https://github.com/GeoscienceAustralia/dea-notebooks/wiki/NotebookTemplate)
- [na] Removed any unused/empty code cells
- [na] Removed any guidance cells (e.g. `General advice`)
- [na] Included relevant tags in the final notebook cell (refer to the [DEA Tags Index](https://knowledge.dea.ga.gov.au/genindex/), and re-use tags if possible)
- [na] Tested notebook on the [DEA Sandbox](https://knowledge.dea.ga.gov.au/guides/setup/Sandbox/sandbox/)
- [na] Cleared all outputs, run notebook from start to finish, and save the notebook in the state where all cells have been sequentially evaluated
- [na] If applicable, update the `Notebook currently compatible with` line below the notebook title to reflect the environments the notebook is compatible with
- [na] Check for any spelling mistakes using the DEA Sandbox's built-in spellchecker (double click on markdown cells then right-click on pink highlighted words). For example:

sandbox_spellchecker

### DIFF
--- a/Tools/dea_tools/classification.py
+++ b/Tools/dea_tools/classification.py
@@ -533,6 +533,9 @@ def _get_training_data_for_shp(
     # Use input feature function and run checks on output
     data = feature_func(dc_query)
 
+    # infer coordinate names
+    x_name, y_name = _find_xy_coords(data)
+
     if not isinstance(data, (xr.Dataset, xr.DataArray)):
         raise TypeError("feature_func must return xarray Dataset or DataArray")
 
@@ -549,10 +552,7 @@ def _get_training_data_for_shp(
         mask = xr_rasterize(dff, data)
         data = data.where(mask)
 
-    if return_coords:
-        # infer coordinate names
-        x_name, y_name = _find_xy_coords(data)
-    
+    if return_coords:    
         # turn coords into variables
         data["x_coord"] = data[x_name] + 0 * data[y_name]
         data["y_coord"] = data[y_name] + 0 * data[x_name]
@@ -574,10 +574,7 @@ def _get_training_data_for_shp(
 
     elif zonal_stats in ["mean", "median", "max", "min"]:
         method_to_call = getattr(data, zonal_stats)
-        if 'x' in data.coords and 'y' in data.coords:
-            stacked = method_to_call(["x", "y"])  # will keep time as dim if present
-        else: 
-            stacked = method_to_call(["longitude", "latitude"]) 
+        stacked = method_to_call([x_name, y_name])  # will keep time as dim if present
         stacked = stacked.to_dataframe().reset_index(drop=True)
         stacked[field] = row[field]
 
--- a/Tools/dea_tools/classification.py
+++ b/Tools/dea_tools/classification.py
@@ -514,8 +514,8 @@ def _get_training_data_for_shp(
             data["x_coord"] = data.x + 0 * data.y
             data["y_coord"] = data.y + 0 * data.x   
         else:
-            data["x_coord"] = data.longitude
-            data["y_coord"] = data.latitude
+            data["x_coord"] = data.longitude + 0 * data.latitude
+            data["y_coord"] = data.latitude + 0 * data.longitude 
 
     # append ID measurement to dataset for tracking failures
     band = list(data.data_vars)[0]
@@ -533,7 +533,10 @@ def _get_training_data_for_shp(
 
     elif zonal_stats in ["mean", "median", "max", "min"]:
         method_to_call = getattr(data, zonal_stats)
-        stacked = method_to_call(["x", "y"])  # will keep time as dim if present
+        if 'x' in data.coords and 'y' in data.coords:
+            stacked = method_to_call(["x", "y"])  # will keep time as dim if present
+        else: 
+            stacked = method_to_call(["longitude", "latitude"]) 
         stacked = stacked.to_dataframe().reset_index(drop=True)
         stacked[field] = row[field]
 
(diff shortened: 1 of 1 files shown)


cbur24 (approved):

robbibt:
Hey @cbur24, @GL-S, apologies - a little slow to the party on this one. Not sure if it's helpful here, but there's also a handy `.spatial_dims` method from `odc-geo` that can be used to extract x and y coordinate names too, e.g.:

image

image

GL-S:
Ah nice, Thanks @robbibt . Chad thought there might be a method already, but I couldn't find it.
Do you know if this always returns y and x in this order, no matter how the dimensions were built in the data array?
E.g. if the data array/dataset has dimensions (time, x, y), would is still output `('y','x')`?

GL-S:
**Ok, I tested it and it does always return them in the same order (y and then x)
e.g.:**
`import xarray as xr
from odc.geo.xr import spatial_dims  

ds = xr.Dataset(
    data_vars={
        "band1": (("time", "x", "y"), [[[1, 2], [3, 4]]])
    },
    coords={
        "time": ["2020-01-01"],
        "x": [10.0, 9.0],
        "y": [100.0, 101.0]
    }
)

dims = spatial_dims(ds)
print(dims) # Output: ('y', 'x')`

**However, it is not robust for unusual names:**

`import xarray as xr
from odc.geo.xr import spatial_dims  

ds = xr.Dataset(
    data_vars={
        "band1": (("time", "latitude_y", "longitude_x"), [[[1, 2], [3, 4]]])
    },
    coords={
        "time": ["2020-01-01"],
        "latitude_y": [10.0, 9.0],
        "longitude_x": [100.0, 101.0]
    }
)

dims = spatial_dims(ds)
print(dims) # OUTPUT--> None`

**The keyword approach might be more robust actually, but those cases where this matters will be super rare...**

robbibt:
> **Ok, I tested it and it does always return them in the same order (y and then x) e.g.:** `import xarray as xr from odc.geo.xr import spatial_dims
> 
> ds = xr.Dataset( data_vars={ "band1": (("time", "x", "y"), [[[1, 2], [3, 4]]]) }, coords={ "time": ["2020-01-01"], "x": [10.0, 9.0], "y": [100.0, 101.0] } )
> 
> dims = spatial_dims(ds) print(dims) # Output: ('y', 'x')`
> 
> **However, it is not robust for unusual names:**
> 
> `import xarray as xr from odc.geo.xr import spatial_dims
> 
> ds = xr.Dataset( data_vars={ "band1": (("time", "latitude_y", "longitude_x"), [[[1, 2], [3, 4]]]) }, coords={ "time": ["2020-01-01"], "latitude_y": [10.0, 9.0], "longitude_x": [100.0, 101.0] } )
> 
> dims = spatial_dims(ds) print(dims) # OUTPUT--> None`
> 
> **The keyword approach might be more robust actually, but those cases where this matters will be super rare...**

It should always return them in the same order. That functionality is linked directly to the `.odc.geobox` of the data, which should always be present if the data was loaded via either Datacube or `odc-stac`, or if the user has `odc.geo.xr` imported. So it's a more standard/interoperable way of recording spatial info.

It's not a big deal either way, but I think my inclination would probably be to actually not support rarer names, because those names aren't likely to be compatiable with `odc-geo` and other ODC-adjacent packages anyway (e.g. `odc-geo` won't be able to automatically create a `geobox` for spatial coordinates with non-standard names). So perhaps, just raise an error instead and nudge the user into choosing more standard options.

GL-S:
Even lat/long returns None

Happy to revert the custom function and use the odc.geo method if simplicity is preferred.
But the method is less robust... would it be possible to edit and improve odc.geo.spatial_dims maybe?

robbibt:
> Even lat/long returns None
> 
> Happy to revert the custom function and use the odc.geo method if simplicity is preferred. But the method is less robust... would it be possible to edit and improve odc.geo.spatial_dims maybe?

It's more of an inherent Open Data Cube restriction: I *think* the only valid x and y coordinate names in the ODC ecosystem are "x"/"y", and "longitude"/"latitude". Everything else isn't going to play nicely with other ODC functionality.

For example, I think `print(dims) # OUTPUT--> None` is actually telling you that `odc-geo` is trying and failing to create a Geobox for that data... which means most other `odc-geo` functionality that requires a Geobox will fail as well (e.g. exporting a GeoTIFF with `ds.odc.write_cog()`, intertactive plotting with `ds.odc.explore()`, reprojecting with `ds.odc.xr_reproject()` etc). I imagine if you do `ds.odc.geobox` you get None too?

GL-S:
@cbur24  found the code (thanks Chad!)  It checks if the dims names are in a list of possible options
https://github.com/opendatacube/odc-geo/blob/9a52c52ba3b0cbddc365454e2c60fbe28b45fc4c/odc/geo/_xr_interop.py#L196

Yeah no problem, I'll draft a PR soon to use the odc.geo method, and make some more testing tomorrow to see no bugs are introduced

PR #1485, opened 2026-04-20
use_odc-geo

GL-S:
A custom function previously added to find spatial dimensions names was removed, and the odc.geo.spatial_dims method is used instead.


robbibt (approved):
This looks fantastic, thanks @GL-S!

GL-S:
Ok, now I have tested it with the code that I used to find the issue the first time and it works well, so I think it can be merged

PR #1486, opened 2026-04-29
Updates to plot_wo

GL-S:
This PR includes updates to the `plot_wo` function.

Previously, the function would assign the colour blue to pixels with the non‑contiguous flag == True. 
This occurred due to the colour‑map logic, which relied on bounds and applied colours to all values indiscriminately.

The core logic of the function was not changed, but a mask is now applied beforehand to remove non‑contiguous pixels.

Also, a new Terrain Shadow class has been added to the legend, and the High Slope + Cloud class was removed (it is now included within the Cloud class).

Finally, the code has been simplified by using a single mapping dictionary that groups labels, bounds and RGB values together.

Example:
BEFORE
image
AFTER
image

### Checklist

- [ ] Removed any unused Python packages from `Load packages`
- [ ] Removed any unused/empty code cells
- [ ] Removed any guidance cells (e.g. `General advice`)
- [ ] Included relevant tags in the final notebook cell (refer to the [DEA Tags Index](https://knowledge.dea.ga.gov.au/genindex/), and re-use tags if possible)
- [ ] If applicable, update the `Notebook currently compatible with` line below the notebook title to reflect the environments the notebook is compatible with
- [ ] Check for any spelling mistakes using the DEA Sandbox's built-in spellchecker (double click on markdown cells then right-click on pink highlighted words). For example:

sandbox_spellchecker


GL-S:
the first commit `binary-dea-wo-plot` was accidental, it was meant to be for another repository, so it accidentally included some testing notebook cells that were later deleted

GL-S:
Dependency to `datacube` was reduced. Although not completely removed because `flags_definition` is still a datacube attribute. However, it should make it a bit easier to adapt the code if we were to move away from datacube in the future

cbur24 (approved):
LGTM @GL-S. Any idea why the tests are failing?

GL-S:
@cbur24 , I don't know why recently any PR I make fails, but it's always on things that should have nothing to do with my changes, so I am not sure why it is happening
image

PR #1505, opened 2026-09-03
add minimum sample size to stratified_random strategy in xr_random_sa…

jennaguffogg:
### Proposed changes
Branch named incorrectly - this is an update to the `xr_random_sampling` function in `validation.py`.

Added a minimum sample size as an optional input to xr_random_sampling, to be used when using the "stratified_random" sampling strategy.

"Stratified_random" distributes sample points between classes based on proportional size - which sometimes leads to very rare/small classes receiving insufficient samples (for example, to perform accuracy assessments on). This allows the user to set a minimum sample count (as an integer). If a class does not have enough pixels to fulfil the request, a warning message is still included to inform the user.

No notebooks are included in this update, though a local notebook was used to test the updated function.

### Checklist

- [ ] Checked the structure of the notebook follows our [DEA-notebooks template](https://github.com/GeoscienceAustralia/dea-notebooks/wiki/NotebookTemplate)
- [ ] Removed any unused Python packages from `Load packages`
- [ ] Removed any unused/empty code cells
- [ ] Removed any guidance cells (e.g. `General advice`)
- [ ] Ensured that all code cells follow the [PEP8 standard](https://www.python.org/dev/peps/pep-0008/) for code. The `jupyterlab_code_formatter` tool can be used to format code cells to a consistent style: select each code cell, then click `Edit` and then one of the `Apply X Formatter` options (`YAPF` or `Black` are recommended).
- [ ] Included relevant tags in the final notebook cell (refer to the [DEA Tags Index](https://knowledge.dea.ga.gov.au/genindex/), and re-use tags if possible)
- [ ] Tested notebook on the [DEA Sandbox](https://knowledge.dea.ga.gov.au/guides/setup/Sandbox/sandbox/)
- [ ] Cleared all outputs, run notebook from start to finish, and save the notebook in the state where all cells have been sequentially evaluated
- [ ] If applicable, update the `Notebook currently compatible with` line below the notebook title to reflect the environments the notebook is compatible with
- [ ] Check for any spelling mistakes using the DEA Sandbox's built-in spellchecker (double click on markdown cells then right-click on pink highlighted words). For example:


Tools/dea_tools/validation.py
```diff
@@ -131,6 +132,9 @@ def xr_random_sampling(
         If setting sampling to 'manual', the provide a dictionary
         of type {'class': numofpoints} mapping the number of points
         to generate for each class.
+    min_sample_size: int, optional
```
robbibt: The doc strings need a space after "min_sample_size"

Tools/dea_tools/validation.py
```diff
@@ -236,6 +240,16 @@ def xr_random_sampling(
...
+            # ensure minimum sample size per class
+            if min_sample_size is not None:
+                for cls in class_sample_sizes:
+                    if class_sample_sizes[cls] < min_sample_size:
+                        if verbose:
+                            print(
+                                f"Class {cls}: increasing sample size from {class_sample_sizes[cls]} to {min_sample_size}."
+                            )
+                        class_sample_sizes[cls] = min_sample_size
+
```
robbibt, line 257: Looks great - if `class_sample_sizes` is a dict, you could potentially simplify a little like this:

```
# Ensure minimum sample size per class
if min_sample_size is not None:
    for cls, count in class_sample_sizes.items():
        if count < min_sample_size:
            if verbose:
                print(
                    f"Class {cls}: increasing sample size from {count} to {min_sample_size}."
                )
            class_sample_sizes[cls] = min_sample_size
```

robbibt:
Looks great @jennaguffogg - if you fix the docstring spacing I can approve the PR; the other loop comment is completely optional. 

Another thing is that it might be nice to add a test for this new param here, checking that the current number of obs are returned. But optional too! 

https://github.com/GeoscienceAustralia/dea-notebooks/blob/develop/Tests%2Fdea_tools%2Ftest_validation.py#L24

robbibt:
@jennaguffogg Something weird is happening with the reviews (it keeps accidently removing people). Can you tag me for review again so I can ✅ ?

robbibt (approved):
Looks fantastic, thanks @jennaguffogg!

